PR #58, opened 2025-12-28
Disable isolation in Miri in order to fix the `ub-detection.yml`

### LINKED ISSUE #54: Fix miri and enable action
Miri failed because the contract broke the isolation

https://github.com/NethermindEth/stellar-private-transactions/actions/runs/20376235759/job/58555366866

We can either:

- disable isolation
- avoid filesystem I/O in tests / in Drop

NiDimi:
- Miri isolation blocks host syscalls (`statx`). Soroban tests write snapshot files using `Env` which trigger `statx` and the check fails
- Disabling isolation lets Miri call into the host OS for filesystem metadata, so the tests can complete.
- We could modify how `Env` works to stop writing snapshot. But I dont think thats the preferred solution 

Closes #54

### DIFF
--- a/.github/workflows/ub-detection.yml
+++ b/.github/workflows/ub-detection.yml
@@ -22,4 +22,4 @@ jobs:
       - run: |
           rustup +nightly component add miri
           cargo +nightly miri setup
-          MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-strict-provenance" cargo +nightly miri test --lib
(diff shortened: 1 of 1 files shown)


maksimryndin (approved):
If we don't ourselves use any unsafe, we can turn off Miri action if it is painful (I know it is sometime takes too much time). Or run it on demand (change a trigger)

PR #73, opened 2026-01-21
Enable wasm-opt

### LINKED ISSUE #48: Enable optimizations and minifications for production app dist
fix wasm-opt issue
Optimize wasm modules sizes by separating dependencies

Fantoni0:
Closes #48 
Enables wasm-opt.
I couldn't fix it directly during crate compilation. It failed because I couldn't pass the required flags there.
So I had to use a `post-build` hook in `Trunk.toml`. 

In general, the build was reduced by 50Kb, and the times are very similar.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ wasm-witness:
 .PHONY: circuits-build
 circuits-build:
 	@echo "Building circuits (this may take a while)..."
-	BUILD_TESTS=1 cargo build -p circuits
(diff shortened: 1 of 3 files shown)


NiDimi (approved):
LGTM!

maksimryndin (changes requested):
Let's not hurry up here, probably, we just need some time to properly figure out how to do it with trunk

PR #85, opened 2026-02-01
Merkle Optimisation

dkgoutham:
Summary
- Add `insert_at(leaf, index)` to MerkleTree for targeted insertion, replacing a 1024-iteration insert loop in transaction-builder with a single call
- Add `serialize()`/`deserialize()` for MerkleTree with IndexedDB `tree_snapshots` store, so page reloads restore the tree from a snapshot instead of replaying all leaves
- Add `build_from_leaves()` batch constructor that builds the tree in a single bottom-up pass, used when rebuilding from > 100 stored leaves


Fantoni0 (changes requested):
The idea is good 🙌 , but some changes can be made. And we need to ensure it does note break the demo functionality.

maksimryndin:
closing in favor of https://github.com/NethermindEth/stellar-private-payments/pull/148

Issue #71, opened 2026-01-19
Fix DB secret storing


Fantoni0:
### Description
Currently, the user's browser stores the secret keys for notes in the DB. This is great for UX, but has some serious implications:
- IndexedDB is accessible to any JS on the page (XSS risk)
- No encryption at rest by default, and the data persists across different sessions.

However, storing only the derivation path means that every time we need to spend a note, the user needs to sign the spending message again. And if he has multiple accounts, he will need to sign in with each one.
We need to research if there is a middle ground between the 2 approaches.

maksimryndin:
take a look as well to https://github.com/NethermindEth/stellar-private-payments/pull/137

depends on https://github.com/NethermindEth/stellar-private-payments/pull/148

ensure also protection in memory (e.g. Zeroize trait in rust)

consider also nullifiers
check Sqlite db encryption extensions

maksimryndin:
Also ensure secrets are protected in cli and app usages

PR #137, opened 2026-03-07
fix(app): encrypt note secrets at rest and reuse session keys per account

Yue-Zhou1:
Addresses #71.

## Summary

- encrypt note `privateKey` and `blinding` before storing them in IndexedDB using AES-256-GCM via the Web Crypto API; no new npm dependencies
- make `saveNote()` and `importNotes()` fail closed instead of falling back to plaintext when encryption keys are not yet derived
- migrate existing plaintext notes to encrypted storage on wallet connect when keys are available, or lazily on first note read
- keep note exports decrypted so backup files remain portable
- cache derived keys in session memory per account so a single account can make multiple spends in one session without signing again for each transaction, and switching `A -> B -> A` reuses `A`'s cached keys
- clear all cached keys on disconnect
- add regression tests for encrypted note storage, migration-adjacent flows, and session key reuse

## Trade-offs

- Derived keys are cached only in memory, bucketed per account, to avoid repeated prompts for both multiple transactions from one account and `A -> B -> A` account switching.
- Keys are not persisted to IndexedDB, `localStorage`, or `sessionStorage`, so refresh and disconnect still require signing again.
- The accepted risk is that a tab-memory compromise exposes derived keys for all accounts that signed in during the session, not just the active one, until disconnect or reload. Note secrets remain encrypted at rest regardless.

## Test Plan

- `cd app && npm test -- --runInBand js/__tests__/notes-store.test.js`
- `cd app && npm test -- --runInBand js/__tests__/ui.test.js`
- `cd app && npm test -- --runInBand`

### DIFF
--- a/app/js/__tests__/notes-store.test.js
+++ b/app/js/__tests__/notes-store.test.js
@@ -1,0 +1,229 @@
+const base64FromBytes = (bytes) => Buffer.from(Uint8Array.from(bytes)).toString('base64');
+
+describe('notes-store issue regression coverage', () => {
+  let notesStore;
+  let userNotes;
+  let signWalletMessage;
+
+  beforeEach(async () => {
+    jest.resetModules();
+    userNotes = new Map();
+    signWalletMessage = jest.fn().mockResolvedValue({
+      signedMessage: base64FromBytes(new Array(64).fill(9)),
+    });
+
+    jest.doMock('../state/db.js', () => ({
+      __esModule: true,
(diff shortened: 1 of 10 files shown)


maksimryndin:
thanks a lot @Yue-Zhou1 ❤️ 

we will need to close it as the general refactoring was done https://github.com/NethermindEth/stellar-private-payments/pull/148

but hope to reuse your work!

PR #140, opened 2026-03-11
use release profile for circuits and trunk

### LINKED ISSUE #134: Add a command to build `dist` in a release mode
Atm, make build builds circuits in the debug mode and the rest of the code (trunk build) as well.

We need some command or env var to build in a release mode (DRY)

Sweetdevil144:
Fix #134 
Add `--release` tag for :
- `.PHONY: build`
- `.PHONY: circuits-build`

As instructed, avoiding usage of`env var` to avoid complications in codebase later on. Rather utilizing `release` in default make builds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,11 @@
 # change where serve, build, and clean write/read compiled assets.
 DIST_DIR ?= dist
 BUILD_TESTS ?=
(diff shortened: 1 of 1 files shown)


Sweetdevil144:
Pushed. `make` does the job now.

PR #143, opened 2026-03-20
Add crate for unified storage

### LINKED ISSUE #139: Unify storage
### Description
Now for agentic payments we need also a CLI solution prototyped here https://github.com/NethermindEth/stellar-private-payments/pull/138.

So there is a split between a browser environment and the server environment. We would like to unify storage using sqlite (both in a browser via OPFS, e.g. https://github.com/Spxg/sqlite-wasm-rs, and locally)

### Benefit (The "Why")
It will reduce maintenance costs and will keep the implementation DRY.

### Acceptance Criteria
both interfaces (browser and cli/server) use the same storage and storage abstraction

Sweetdevil144:
Fixes #139 

- Need to build up other PRs for Rust on top of this PR


maksimryndin (commented):
Good work!

Let's make changes gradual. Could we have substitute app/js/state/db.js with the in-memory wasm implementation?

And it would just work?

maksimryndin (commented):
I approve it here but not at the PR level.

We will first merge the consecutive pr into this one, then check this branch works and then merge into main, ok?

maksimryndin (changes requested):

maksimryndin:
@Sweetdevil144 before the review
1) let's fix the CI errors
2) let's remove 

> 1. Since we reset to `wasm.rs`, Jest can not load WASM build artifacts during `npm test`, so every WASM module needs a mock. same as existing `__mocks__/prover.js` and `__mocks__/witness.js`
> 2. rusqlite in `native.rs` doesn't compile for `wasm32`. both the files use entirely different SQLite libraries so methods in both differ as well including `open()`
> 3. wasm-bindgen does NOT auto-init. `wasm.js` still requires an explicit init() call to load the .wasm binary
> 4. Fixed. Used git diff to restore comments which were removed by claude in middle of changes
> 
> Should we wrap `native.rs` and `wasm.rs` into a single file. My original thinking was to keep them separate for better readability and separate functionality, additionally, combining them into one would mean hundreds of `#[cfg]` block declarations on functions :)

For example,
Why do we have - can this code be unified?
- https://github.com/NethermindEth/stellar-private-payments/pull/143/changes#diff-4d0b74fa2e88bd396acb398c4906543f9b94f21998a28968450577801c54cef6R274
- https://github.com/NethermindEth/stellar-private-payments/pull/143/changes#diff-a3c431bdbf236b8a2690b8c2b8a8ee42942beabbf97c595078218bf5e65ef1f6R133

```rust
/// Inserts or replaces a nullifier record.
    pub fn put_nullifier(&self, nullifier: &PoolNullifier) -> anyhow::Result<()> {
        let s = self.conn.prepare(
            "INSERT OR REPLACE INTO pool_nullifiers (nullifier, ledger) VALUES (?1, ?2)",
        )?;
        s.bind_text(1, &nullifier.nullifier)?;
        s.bind_i64(2, i64::from(nullifier.ledger))?;
        s.step_row()?;
        Ok(())
    }
```

```rust
pub fn put_nullifier(&self, nullifier: &PoolNullifier) -> anyhow::Result<()> {
        self.conn
            .execute(
                "INSERT OR REPLACE INTO pool_nullifiers (nullifier, ledger) VALUES (?1, ?2)",
                params![&nullifier.nullifier, i64::from(nullifier.ledger)],
            )
            .context("put_nullifier")?;
        Ok(())
    }
```

maksimryndin:
@Sweetdevil144 thanks a lot for the contribution but it won't merged - as the problem was much wider (as I recognized), so the one will follow https://github.com/NethermindEth/stellar-private-payments/pull/148

Issue #170, opened 2026-04-17
Unify usage of types::Field in the prover crate


maksimryndin:
General principle - conversion to ark-works Scalar should happen at the last moment when needed

App-local datastructures (like MerklePrefixTree, Prover etc) should rely on types::Field.

Also in flows.rs - fix encoding of field elements to circuit inputs using convenient methods from types::Field

related https://github.com/NethermindEth/stellar-private-payments/issues/154
depends on https://github.com/NethermindEth/stellar-private-payments/pull/148

uttkarzsh:
Hi. I'm planning to work on this issue.

I'll also be keeping an eye on the related issue https://github.com/NethermindEth/stellar-private-payments/issues/154 to ensure compatibility with any structural changes.

maksimryndin:
closed with https://github.com/NethermindEth/stellar-private-payments/pull/179

PR #148, opened 2026-04-16
Unified state and Rust app core

maksimryndin:
# Summary

- [x] the state now is preserved in Sqlite (to make a single unified storage across platforms)
- [x] most of the core logic now is in Rust (again for ease of cross platform support)
- [x] js is a thin UI layer + interaction with wallet and the tx simulation and submission
- [x] indexer is a background process running while the tab is open - see https://github.com/NethermindEth/stellar-private-payments/blob/feature-unified-state/app/ARCHITECTURE.md
- [x] the chain is a single source of truth - no optimistic save
- [x] the legal disclaimer is added as well at the core level and per account and disclaimer text hash
- [x] new types are introduces for keys, amounts etc to enforce strict type checking
- [x] ci passes

closes https://github.com/NethermindEth/stellar-private-payments/issues/139 https://github.com/NethermindEth/stellar-private-payments/issues/136

# How to review

The PR is huge, sorry for that. But due to tight and wide penetration of Rust (via wasm-bindgen) the incremental approach didn't work (see https://github.com/NethermindEth/stellar-private-payments/pull/143). 
So
1. Refresh project structure https://github.com/NethermindEth/stellar-private-payments/blob/feature-unified-state/CONTRIBUTING.md#project-structure
2. Get familiar with the architecture https://github.com/NethermindEth/stellar-private-payments/blob/feature-unified-state/app/ARCHITECTURE.md
3. Some things to pay special attention to: 
- deposit and withdraw amounts are represented with ExtAmount(i128) because anyway Soroban token transfer accepts only i128.
- note amounts are NoteAmount(u64) because note is encrypted with 8 bytes for the amount.
- a separate Field type is introduced to have types crate lightweight and introduce convenient conversion and serialization traits. And be agnostic of a specific field instantiation
- do we really need `u64` for the leaf index in ASP tree - for https://github.com/NethermindEth/stellar-private-payments/pull/148/changes#diff-32304f6136c0b41bef6008a803b45e4d949bf71a4097cceaa8a609a8ee992d30R158  (in the commitments tree even which is probably greater - we have `u32`)
- app/crates/core/prover/src/flows.rs
4. JS part was AI-regenerated to integrate - so probably just running it in the browser and testing manually makes sense. Also js tests (as they actually didn't test too much because of mocks) were removed - see also https://github.com/NethermindEth/stellar-private-payments/issues/168

PR #171, opened 2026-04-19
refactor: unify usage of types::Field in the prover crate

uttkarzsh:
Implements the changes described in issue https://github.com/NethermindEth/stellar-private-payments/issues/170

Summary of changes:
Structures now use `types::Field`  and use arkworks `Scalar` type only when needed for hashing. Also updated the Merkle Tree structures as well to use Field instead of Scalar.
Used standard circuit input encoding using `Field::to_le_bytes()`

- cargo test -p prover --tests  (all passed)
- cargo clippy passed
- cargo fmt applied

continued from pr https://github.com/NethermindEth/stellar-private-payments/pull/148

### DIFF
--- a/app/crates/core/prover/src/merkle.rs
+++ b/app/crates/core/prover/src/merkle.rs
@@ -6,44 +6,58 @@
 use alloc::vec::Vec;
 
 use anyhow::{Result, anyhow};
-use zkhash::fields::bn256::FpBN256 as Scalar;
+use zkhash::{ark_ff::PrimeField, fields::bn256::FpBN256 as Scalar};
 
-use crate::{
-    crypto,
-    serialization::{bytes_to_scalar, scalar_to_bytes},
-};
-use types::Field as AppField;
+use crate::{crypto, serialization::scalar_to_bytes};
+use types::Field;
 
 // Re-export core merkle functions from circuits
 pub use circuits::core::merkle::{
     merkle_proof as merkle_proof_internal, merkle_root, poseidon2_compression,
 };
 
+fn field_to_scalar(field: Field) -> Scalar {
+    Scalar::from_le_bytes_mod_order(&field.to_le_bytes())
+}
+
+fn scalar_to_field(scalar: Scalar) -> Field {
+    let le = scalar_to_bytes(&scalar);
+    let le: [u8; 32] = le.try_into().expect("scalar bytes length");
+    Field::try_from_le_bytes(le).expect("scalar to field conversion")
+}
+
+fn hash_pair(left: Field, right: Field) -> Field {
+    let left_s = field_to_scalar(left);
+    let right_s = field_to_scalar(right);
+    let hashed = poseidon2_compression(left_s, right_s);
+    scalar_to_field(hashed)
+}
+
 /// Merkle proof data
 pub struct MerkleProof {
     /// Path elements
-    pub path_elements: Vec<AppField>,
+    pub path_elements: Vec<Field>,
     /// Path indices as a single scalar
-    pub path_indices: AppField,
+    pub path_indices: Field,
     /// Computed root
-    pub root: AppField,
+    pub root: Field,
     /// Number of levels
     pub levels: usize,
 }
 
 impl MerkleProof {
     /// Get path elements, one field element per level.
-    pub fn path_elements(&self) -> Vec<AppField> {
+    pub fn path_elements(&self) -> Vec<Field> {
         self.path_elements.clone()
     }
 
     /// Get path indices packed into a field element.
-    pub fn path_indices(&self) -> AppField {
+    pub fn path_indices(&self) -> Field {
         self.path_indices
     }
 
     /// Get computed root as a field element.
-    pub fn root(&self) -> AppField {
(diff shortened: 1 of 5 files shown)


maksimryndin:
hey @uttkarzsh ! ✋  Thanks a lot for the contribtuion!
Could we also please make sure https://github.com/NethermindEth/stellar-private-payments/blob/feature-unified-state/app/crates/core/prover/src/flows.rs#L31 and https://github.com/NethermindEth/stellar-private-payments/blob/feature-unified-state/app/crates/core/prover/src/flows.rs#L75 to use `types::Field` instead of plain `[u32; 32]`? where the field element is expected?

app/crates/platforms/web/src/workers/storage.rs
```diff
@@ -636,6 +636,11 @@ fn build_pool_inputs(
             })?;
 
         let (path_elements, path_indices) = tree.proof_bytes(leaf_index)?;
```
maksimryndin: Maybe we should change `proof_bytes` signature to return Vec<Field> and Field instead of arrays (probably the method name should also be changed then)? wdyt @uttkarzsh ?

If that method used only here, then probably it makes sense?

app/crates/platforms/web/src/workers/prover.rs
```diff
@@ -178,17 +178,11 @@ fn prove_from_artifacts(transact_artifacts: TransactArtifacts) -> Result<Prepare
         }
 
         let p = transact_artifacts.prepared.clone();
```
maksimryndin: A small improvement we can introduce here (I didn't do that initially) - probably, we don't need clone here - we can deconstruct the struct. And then only clone `ext_data` earlier.

app/crates/core/prover/src/sparse_merkle.rs
```diff
@@ -403,24 +419,24 @@ impl WasmSparseMerkleTree {
...
         let find_result = self.inner.find(&key).map_err(|e| anyhow!(e))?;
 
         // Pad siblings to max_levels
         let mut siblings = find_result.siblings.clone();
         while siblings.len() < max_levels {
-            siblings.push(Scalar::from(0u64));
+            siblings.push(Field::ZERO);
         }
 
         Ok(WasmSMTProof {
```
maksimryndin: Let's rename this struct to just `SMTProof`

app/crates/core/prover/src/merkle.rs
```diff
@@ -6,44 +6,58 @@
...
-use types::Field as AppField;
+use crate::{crypto, serialization::scalar_to_bytes};
+use types::Field;
 
 // Re-export core merkle functions from circuits
 pub use circuits::core::merkle::{
     merkle_proof as merkle_proof_internal, merkle_root, poseidon2_compression,
 };
 
+fn field_to_scalar(field: Field) -> Scalar {
```
maksimryndin, line 19: could we add a few simple unit tests here to check the conversion at edge cases? (e.g. at the modulo p - 1, at modulo p, at 0 etc)?

maksimryndin (commented):
Looks good! 👍  thanks a lot @uttkarzsh !

some nit picks!

app/crates/platforms/web/src/workers/storage.rs
uttkarzsh: Changed the call from proof_bytes() to proof() which already returns Vec and Field that we need. If we change the signature, we will be kind of duplicating what proof() already does for us. And proof_bytes is only left in the test cases at the moment, so i kept it as-is for now as a byte-level helper if needed in the future (maybe). If you’d like I can remove it entirely and update the test cases. :)

app/crates/platforms/web/src/workers/prover.rs
uttkarzsh: I tried destructuring, but prepared is a PreparedTx (not PreparedTxPublic) and the field names differ (public_amount vs public_amount_field), so I kept the explicit mapping from PreparedTx → PreparedTxPublic to avoid type/field mismatches. I still removed the clone()   and now clone ext_data up front. cargo check -p web --all-features passes. lmk there is something else that you’d like me to do :)

app/crates/core/prover/src/sparse_merkle.rs
uttkarzsh: Done 👍

app/crates/core/prover/src/merkle.rs
uttkarzsh: Done. Tests now cover 0, 1, modulus - 1, and modulus (reduces to 0).

app/crates/platforms/web/src/workers/storage.rs
maksimryndin: let's remove proof_bytes then if don't use it anywhere else except tests?

uttkarzsh: Done :)

maksimryndin:
Hey @uttkarzsh! ✋ 

Sorry, seems GH closed automatically - could you pls rebase you PR on the latest main?

PR #172, opened 2026-04-22
Add timeout to rpc & Increase memory to wasm

### LINKED ISSUE #157: Increase WASM memory to future proof 4GB
by default the available memory limit baked into wasm and it is usually less than available

### LINKED ISSUE #163: Add timeout for JSON RPC in wasm
see `app/crates/core/stellar/src/rpc.rs`

Sweetdevil144:
1. Fixes #163 : Added default timeout to RPC, default = 30 seconds : `const RPC_TIMEOUT_SECS: u32 = 30;`
2. Fixes #157 : without `--max-memory`, the default behavior caps at ~**2GB** in practice (browser-imposed), which is "less than available" of what we require. Capped memory to 4GB (this should be enough for entire WASM modules), with default as 64MB


maksimryndin:
@Sweetdevil144 could you pls rebase on the main?

maksimryndin:
@Sweetdevil144 could we introduce a timeout test? (ideally for both targets)

maksimryndin:
@codex review

app/crates/core/stellar/src/rpc.rs
```diff
@@ -194,10 +196,21 @@ pub struct SimulateTransactionResponse {
 pub struct Client {
     base_url: String,
     http_client: reqwest::Client,
+    /// Read by the WASM timeout path
+    #[cfg_attr(not(target_arch = "wasm32"), allow(dead_code))]
```
maksimryndin: the comment, the use vs the `cfg` attr seems to contradict each other, no?

app/crates/core/stellar/src/rpc.rs
```diff
@@ -244,14 +258,31 @@ impl Client {
...
+            self.http_client
+                .post(&self.base_url)
+                .json(&payload)
+                .send()
+                .await?
+                .json::<JsonRpcResponse<R>>()
+                .await
+        };
+
+        #[cfg(target_arch = "wasm32")]
```
maksimryndin, line 270: it would be nice to test this branch rather than native one (as the native one uses a standard timeout mechanism from the dependency), while here the timeout is re-implemented

app/crates/core/stellar/src/rpc.rs
Sweetdevil144: Resolved. Native build no longer carries the field at all

app/crates/core/stellar/src/rpc.rs
Sweetdevil144: Replaced with two #[wasm_bindgen_test] cases against the extracted race_with_timeout helper. Accordingly CI changes made and tested. Native test removed.

maksimryndin:
@Sweetdevil144 looks good! let's merge the latest main and fix the conflicts?

Sweetdevil144:
Resolving...

Sweetdevil144:
@maksimryndin conflicts resolved

maksimryndin (approved):
LGTM! 👍 
Thanks a lot @Sweetdevil144 ! 👍

PR #178, opened 2026-04-23
Unify usage of types::Field in the prover crate

uttkarzsh:
Implements the changes described in the issue https://github.com/NethermindEth/stellar-private-payments/issues/170

Summary of changes:
- Predominantly use types::Field instead of raw bytes and arkworks Scalar. Make use of arkworks Scalar type only when needed for hashing. Also updated the Merkle Tree structures as well to use Field instead of Scalar.
- Used standard circuit input encoding using Field::to_le_bytes().
- Remove proof_bytes() method as it is not used anywhere, only in tests.
- Added a few more tests for edge cases like modulo p, p-1, 0, 1 etc. for field_to_scalar() conversion


maksimryndin:
hey @uttkarzsh ! ✋ thanks for the contribution! could you please rebase on the latest main and ensure that only your changes are in the diff?

uttkarzsh:
hey @maksimryndin 👋🏻 ! Apologies. Opened a new pr with only my changes. Check https://github.com/NethermindEth/stellar-private-payments/pull/179

PR #179, opened 2026-04-23
unify usage of types::Field in the prover crate

uttkarzsh:
Implements the changes described in the issue https://github.com/NethermindEth/stellar-private-payments/issues/170

Summary of changes:
- Predominantly use types::Field instead of raw bytes and arkworks Scalar. Make use of arkworks Scalar type only when needed for hashing. Also updated the Merkle Tree structures as well to use Field instead of Scalar.
- Used standard circuit input encoding using Field::to_le_bytes().
- Remove proof_bytes() method as it is not used anywhere, only in tests.
- Added a few more tests for edge cases like modulo p, p-1, 0, 1 etc. for field_to_scalar() conversion

### DIFF
--- a/app/crates/core/prover/src/flows.rs
+++ b/app/crates/core/prover/src/flows.rs
@@ -33,11 +33,11 @@ pub struct TransactInputNote {
     pub amount_stroops: NoteAmount,
     /// Note blinding factor as a BN254 scalar field element.
     pub blinding: Field,
-    /// Merkle proof sibling hashes as BN254 scalars (little-endian field
-    /// bytes), one element per tree level.
-    pub merkle_path_elements: Vec<[u8; 32]>,
-    /// Merkle path indices packed into a scalar (little-endian field bytes).
-    pub merkle_path_indices: [u8; 32],
+    /// Merkle proof sibling hashes as BN254 field elements, one element per
+    /// tree level.
+    pub merkle_path_elements: Vec<Field>,
+    /// Merkle path indices packed into a BN254 field element.
+    pub merkle_path_indices: Field,
 }
 
 /// Output note specification for a pool transaction.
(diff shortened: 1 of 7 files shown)


app/crates/core/prover/src/flows.rs
```diff
@@ -771,13 +769,22 @@ fn dummy_input(tree_depth: usize) -> Result<TransactInputNote> {
...
     })
 }
 
-fn note_amount_to_field_le(amount: NoteAmount) -> [u8; 32] {
-    Field::from(amount).to_le_bytes()
+fn note_amount_to_field(amount: NoteAmount) -> Field {
+    Field::from(amount)
+}
+
+fn field_to_circuit_hex(field: Field) -> Result<String> {
```
maksimryndin: nit pick: this function can take by reference - no need to take ownership over the value

app/crates/core/prover/src/flows.rs
```diff
@@ -771,13 +769,22 @@ fn dummy_input(tree_depth: usize) -> Result<TransactInputNote> {
...
+    field_bytes_to_hex(&field.to_le_bytes())
```
maksimryndin, line 782: Here we could use `types::Field` native hex formatting but probably fine for now

maksimryndin (approved):
Looks good! Thanks @uttkarzsh ! 👍

app/crates/core/prover/src/flows.rs
maksimryndin: let's keep as-is for this PR

maksimryndin:
@uttkarzsh you can use `.githooks/pre-push` to check the flows

uttkarzsh:
hey @maksimryndin ! I updated the flow helper to borrow instead of take ownership as you suggested and ran .githooks/pre-push for end to end test. all checks passed. i also updated the  hook to use the new stellar contract build --profile release flag as the cli no longer accepts --optimize flag.

.githooks/pre-push
```diff
@@ -48,7 +48,7 @@ fi
 
 mkdir -p target/stellar
 for pkg in asp-membership asp-non-membership circom-groth16-verifier pool; do
-    if ! stellar contract build --manifest-path Cargo.toml --out-dir target/stellar --optimize --package "$pkg"; then
+    if ! stellar contract build --manifest-path Cargo.toml --out-dir target/stellar --profile release --package "$pkg"; then
```
maksimryndin: which version of `stellar` do you use?

uttkarzsh: i have `stellar 21.3.0 (v20.0.0-249-g11746937e813346272faf68e70cbdbc1b0df780f)`

maksimryndin: It is probably an old one. Could you pls try to update to the latest one and check again that flag?

uttkarzsh: i updated stellar to the latest. and --optimize works now. I'll change it back to that.

maksimryndin:
@uttkarzsh all good, we just need signed commits https://github.com/NethermindEth/stellar-private-payments/blob/main/CONTRIBUTING.md#commit-signing, you can rebase on the latest main and squash your changes into a single signed commit

uttkarzsh:
hey @maksimryndin 👋🏻 Rebasing done on latest main, all changes squashed into one signed commit 😁.

app/crates/core/prover/src/merkle.rs
```diff
@@ -6,44 +6,58 @@
...
-use types::Field as AppField;
+use crate::{crypto, serialization::scalar_to_bytes};
+use types::Field;
 
 // Re-export core merkle functions from circuits
 pub use circuits::core::merkle::{
     merkle_proof as merkle_proof_internal, merkle_root, poseidon2_compression,
 };
 
+fn field_to_scalar(field: Field) -> Scalar {
```
maksimryndin: Here we can accept as well by reference and it is duplicated in `sparse_merkle.rs` - maybe we can unify and put in the conversions.rs?

uttkarzsh: Done!!

maksimryndin:
awesome work @uttkarzsh 🚀 thanks a lot!

PR #184, opened 2026-04-24
(chore) fix when user doesnt accept terms

maksimryndin:
Current behaviour: If a user doesn't accept disclaimer terms on the onboading, then clicks connect wallet, then again decline - we see the error in console. It doesn't prevent a user from onboarding again, just we don't need to have an error in console.

### DIFF
--- a/app/js/ui/onboarding-wizard.js
+++ b/app/js/ui/onboarding-wizard.js
@@ -3,6 +3,24 @@ import { deriveKeysFromWallet } from '../wallet.js';
 
 const STORAGE_PERSIST_FLAG = 'poolstellar_storage_persist_prompted';
 
+class OnboardingCancelledError extends Error {
+    constructor(message = 'Onboarding cancelled') {
+        super(message);
+        this.name = 'OnboardingCancelledError';
+    }
+}
+
+class TermsDeclinedError extends Error {
+    constructor(message = 'Terms & Conditions must be accepted to use this service.') {
+        super(message);
+        this.name = 'TermsDeclinedError';
+    }
+}
+
+export function isOnboardingCancelledError(e) {
+    return !!e && (e.name === 'OnboardingCancelledError' || e instanceof OnboardingCancelledError);
+}
+
 function hasStorageManager() {
     return (
         typeof navigator !== 'undefined' &&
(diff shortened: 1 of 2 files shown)


Fantoni0 (commented):
In general it works 🙌 
But the first time the terms and conditions pop up (and the auto flag is true) I don't see the toast notification (Onboarding cancelled) if I reject the terms. However if I click again on connect wallet and decline or abort a 2nd time, the notification is shown.

Not sure if this is the desired behavior, but I think it's related to the auto flag and the branching `if (!auto)` before showing the toast in the initial connection promise.

app/js/ui/onboarding-wizard.js
```diff
@@ -3,6 +3,24 @@ import { deriveKeysFromWallet } from '../wallet.js';
...
 const STORAGE_PERSIST_FLAG = 'poolstellar_storage_persist_prompted';
 
+class OnboardingCancelledError extends Error {
+    constructor(message = 'Onboarding cancelled') {
+        super(message);
+        this.name = 'OnboardingCancelledError';
+    }
+}
+
+class TermsDeclinedError extends Error {
```
NiDimi, line 13: Would here be better to extend `OnboardingCancelledError`? This means that in `isOnboardingCancelledError` function would also catch the termsDeclinedError

app/js/ui/onboarding-wizard.js
```diff
@@ -327,7 +347,7 @@ export async function runOnboardingWizard({ address, setButtonLoading } = {}) {
             renderActions([decline, accept]);
         });
 
-        if (cancelled) throw new Error('Onboarding cancelled');
+        if (cancelled) throw new OnboardingCancelledError();
```
NiDimi, line 350: Its just a nit pick and mostly checking my knowledge, but why do we need the cancelled? 

- onAbort we already reject with the error
- on Decline we reject with the error
- on accept obviously is not needed.

In what scenario would the cancelled check trigger?

NiDimi (commented):
Same behavior as Antonio. Also left some nit picks

maksimryndin:
all ui changes will be introduced in a separate PR, no need to keep it

PR #200, opened 2026-05-13
feat(circom-groth16-verifier): embed VK at compile time, remove storage (#196)

### LINKED ISSUE #196: Hardcode Groth16 verification key in verifier contract instead of passing it at deployment
We should stop passing the Groth16 verification key to the verifier contract at deployment time.

  Instead, we need tooling that generates or builds a circuit-specific verifier contract from a selected VK. The resulting verifier WASM should already contain the VK, so deployment only deploys the verifier contract and does not initialize/set the VK.

 If the VK is embedded in the verifier WASM at compile time:

  - there is no persistent storage write during initialization
  - there is no persistent storage read during `verify`
  - verification uses static contract data instead of contract storage

thinktanktom:
## What's this PR do?

Closes #196

Embeds the Groth16 verification key directly into the contract WASM at
compile time, removing the two storage operations that previously
occurred on every deployment and every proof verification.

---

## Context

Previously the VK was written to persistent ledger storage in a
`__constructor` call at deploy time, then read back on every `verify()`
invocation. Issue #196 asked to eliminate both by baking the key into
the WASM binary itself.

The mechanism is a Cargo build script (`build.rs`) that runs at compile
time. It reads a snarkjs `verification_key.json` from the
`VERIFIER_VK_JSON` environment variable, converts the G1/G2 decimal
string coordinates to big-endian byte arrays via `num-bigint`, and
writes static `const [u8; N]` arrays into a `vk.rs` file in Cargo's
`OUT_DIR`. `lib.rs` pulls these in with a single `include!` macro.

A few non-obvious decisions worth calling out:

**Zero-filled fallback in `build.rs`:** When `VERIFIER_VK_JSON` is
unset (CI, `cargo check`, unit tests), `build.rs` emits zero-filled
arrays rather than failing. This keeps the compilation graph intact
without requiring a real key in every environment. Tests bypass the
embedded key entirely by calling `verify_with_vk` directly with a
locally constructed `VerificationKey`.

**`embedded_vk()` re-constructs on every call:** The function that
materialises compile-time constants into SDK curve types is called
afresh on each `verify()` invocation. Curve-type construction is cheap
relative to the BN254 pairing operations that follow, so this is not a
performance concern in practice — but it is worth knowing there is no
caching.

**`Groth16Error::NotInitialized` removed:** The variant was only ever
returned by the constructor path, which no longer exists. With the VK
baked in there is no initialisation step and no realistic future path
that would need it. It has been removed from `contracts/types/src/lib.rs`
rather than left as dead code.

**Breaking CLI change:** `deploy.sh` no longer accepts `--vk-json` or
`--vk-file`. The verifier must be pre-built with
`scripts/build-verifier-with-vk.sh` before running the deploy script.
Operators with existing CI pipelines or runbooks that pass either flag
will need to update them.

---

## 🔍 Unrelated issue spotted during development

While working on this PR, the `wasmer` git dependency in
`app/crates/core/witness/Cargo.toml` caused repeated pre-push hook
failures due to Cargo cloning the entire wasmer repository to resolve a
single pinned commit. The wasmer repo is large enough that the clone
consistently times out or drops the SSL session on anything but a very
fast connection, blocking any contributor without a pre-warmed Cargo
cache from pushing.

The fix is a one-line change to the workspace `Cargo.toml` — adding
`shallow = true` to the existing git dependency:

```toml
# see https://github.com/NethermindEth/stellar-private-payments/issues/192
wasmer = {
  git = "https://github.com/wasmerio/wasmer",
  rev = "763e9f2800644f51ce27f6f5c1752776da16ddd1",
  shallow = true,
  default-features = false
}
```

This instructs Cargo to fetch only the single pinned commit rather than
the full history, which is the behaviour that was always intended.
The pinned commit, vendored `cranelift-control` patch, and `js`/
`cranelift` feature flags are all left untouched — this is purely a
fetch optimisation with no behavioural change.

Not included in this PR since it touches unrelated infrastructure, but
raising it here so it doesn't get lost. Happy to split it into a
follow-up if useful.

---

### Related issues

Closes #196

---


contracts/circom-groth16-verifier/Cargo.toml
```diff
@@ -14,6 +14,8 @@ contract-types = { workspace = true }
 soroban-sdk = { workspace = true, features = ["alloc"] }
 
 [build-dependencies]
+num-bigint = { version = "0.4", features = ["std"] }
```
maksimryndin: can we use workspace dependencies here?

deployments/scripts/deploy.sh
```diff
@@ -27,8 +27,6 @@ Options:
   --asp-levels N        Merkle tree levels for asp-membership (required)
   --pool-levels N       Merkle tree levels for pool (required)
   --max-deposit U256    Maximum deposit amount (required)
-  --vk-json JSON        VerificationKeyBytes JSON for circom verifier constructor
-  --vk-file PATH        JSON file containing VerificationKeyBytes
```
maksimryndin, line 31: Let's keep this argument to feed into `build-verifier-with-vk`, wdyt?

contracts/circom-groth16-verifier/build.rs
```diff
@@ -0,0 +1,149 @@
...
+        "field element too large ({} bytes): `{s}`",
+        be.len()
+    );
+    let mut out = [0u8; 32];
+    out[32 - be.len()..].copy_from_slice(&be);
+    out
+}
+
+/// Encode a snarkjs G1 point `[x, y, "1"]` as 64 bytes (x || y, big-endian).
+fn g1_bytes(pt: &Value) -> [u8; 64] {
```
maksimryndin, line 40: could we reuse here `circuit_keys::g1_to_soroban_bytes`?

contracts/circom-groth16-verifier/build.rs
```diff
@@ -0,0 +1,149 @@
...
+    out[..32].copy_from_slice(&to_bytes32(x));
+    out[32..].copy_from_slice(&to_bytes32(y));
+    out
+}
+
+/// Encode a snarkjs G2 point `[[x_c1, x_c0], [y_c1, y_c0], ...]` as 128 bytes.
+///
+/// Soroban's `Bn254G2Affine` expects the imaginary (c1) component before the
+/// real (c0) component, which already matches the snarkjs array order.
+fn g2_bytes(pt: &Value) -> [u8; 128] {
```
maksimryndin, line 55: could we reuse here `circuit_keys::g2_to_soroban_bytes`?

maksimryndin (commented):
Hey @thinktanktom ! thanks a lot for the contribution 🚀 

could you please address some comments and let's fix some ci failures?

(on the shallow copy - thanks - feel free to add here as well)

maksimryndin:
@codex review

thinktanktom:
> Hey @thinktanktom ! thanks a lot for the contribution 🚀
> 
> could you please address some comments and let's fix some ci failures?
> 
> (on the shallow copy - thanks - feel free to add here as well)

Sure I'll probably get to it by tomorrow. Hope that's all right.

deployments/scripts/deploy.sh
thinktanktom: Yes I agree. That would be a more efficient approach.

e2e-tests/src/tests/utils.rs
```diff
@@ -17,14 +15,31 @@ use circuits::test::utils::{
...
+/// Mock verifier contract for testing — always approves any proof.
+///
+/// The ZK proof verification logic is covered by the circom-groth16-verifier
+/// unit tests; e2e pool tests only need to exercise the pool's transaction
+/// flow, so a verifier that unconditionally returns `Ok(true)` is sufficient.
+#[contract]
+struct MockVerifier;
+
+#[contractimpl]
+impl MockVerifier {
```
maksimryndin: I am not sure I like the idea to mock the verifier in the e2e test. Probably, we should consider injecting vk_bytes somehow inside the default wasm binary of CircomGroth16Verifier (maybe even mere bytes replacement of a placeholder would work??)

maksimryndin (commented):
Thanks a lot for the contribution @thinktanktom !

Left one comment - I think worth to check if it is possible

e2e-tests/src/tests/utils.rs
thinktanktom: Replaced the `MockVerifier` with the real `CircomGroth16Verifier` in `deploy_contracts`.

To make the VK available at compile time without manual setup, I added a `VERIFIER_VK_JSON` entry in `.cargo/config.toml` pointing to `testdata/policy_tx_2_2_vk.json` (no `force` flag, so CI deployment jobs can still override it with the ceremony key). The VK file itself is now tracked in git via a `.gitignore` exception so `build.rs` can resolve it in CI.

The e2e test now exercises the real verifier end-to-end with no mocking.

contracts/circom-groth16-verifier/build.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    let content = match env::var("VERIFIER_VK_JSON") {
+        Ok(path) => {
+            let path = PathBuf::from(&path);
+            println!("cargo:rerun-if-changed={}", path.display());
+            let json = fs::read_to_string(&path)
+                .unwrap_or_else(|e| panic!("failed to read VK file `{}`: {e}", path.display()));
+            vk_rs_from_json(&json)
+        }
+        Err(_) => placeholder_vk_rs(),
```
maksimryndin: if we now have a default `VERIFIER_VK_JSON`, do we need a placeholder here?

let's remove `placeholder_vk_rs`

```rs
let path = env::var("VERIFIER_VK_JSON").expect("VERIFIER_VK_JSON not set");
let path = PathBuf::from(&path);
println!("cargo:rerun-if-changed={}", path.display());
let json = fs::read_to_string(&path)
     .unwrap_or_else(|e| panic!("failed to read VK file `{}`: {e}", path.display()));
let content = vk_rs_from_json(&json);
```

.gitignore
```diff
@@ -53,3 +54,5 @@ docs/book/
 ceremony/
 *.ptau
 artifacts.zip
+flake.nix
```
maksimryndin: if you have a working nix setup for this repo, feel free to contribute a separate PR :)

_(but here these changes are irrelevant - let's remove)_

e2e-tests/src/tests/utils.rs
maksimryndin: amazing! 🥇

testdata/policy_tx_2_2_vk.json
```diff
@@ -0,0 +1,114 @@
+{
```
maksimryndin: and let's remove this file from the git history (and the .gitignore) - as in any case we need circuits built in the cicd, so this prerequisite is preserved?

.gitignore
```diff
@@ -45,6 +45,7 @@ app/e2e/playwright-report/
 
 # Testdata
 testdata/*
+!testdata/policy_tx_2_2_vk.json
```
maksimryndin: let's remove it as we build in the flows https://github.com/NethermindEth/stellar-private-payments/blob/main/.github/workflows/build-and-test.yml#L30

contracts/circom-groth16-verifier/build.rs
```diff
@@ -0,0 +1,143 @@
...
+    writeln!(
+        out,
+        "const VK_IC: [[u8; 64]; {ic_len}] = [{}];",
+        ic_items.join(",")
+    )
+    .expect("infallible write to String");
+    out
+}
+
+/// Zero-filled placeholder emitted when `VERIFIER_VK_JSON` is not set.
```
maksimryndin: let's remove `placeholder_vk_rs`

maksimryndin (approved):

PR #203, opened 2026-05-18
Feature/selective disclosure circuit

Fantoni0:
It adds circuit support for selective disclosure:

- Adds new selective disclosure circuit.
- Updates build.rs to include additional circuits.
- Adds circom tests for the new circuit

---
Selective disclosure is a feature that allows a user to generate a zk-proof for claiming ownership of a note privately. Basically, claiming they know the private key, amount, and blinding factor behind the note, without spending the note or revealing those secrets. The proof will be tied to a context: pool address, merkle root and authority id to avoid proof replaying.

The selective disclosure allows batched proving for multiple notes, on different merkle roots, but for the SAME context.

### DIFF
--- a/circuits/src/selectiveDisclosure.circom
+++ b/circuits/src/selectiveDisclosure.circom
@@ -1,0 +1,63 @@
+pragma circom 2.2.2;
+
+// Selective Disclosure Circuit
+include "./merkleProof.circom";
+include "./poseidon2/poseidon2_hash.circom";
+include "./keypair.circom";
+
+// Allows a user to demonstrate ownership of a note commitment without revealing its secrets
+// * levels: Number of levels in the Merkle tree that holds the note commitments
+// * nNotes: Number of notes to prove ownership of.
+// Right now we allow proving ownership of multiple notes on different roots BUT for the same external context:
+// - Purpose
+// - Authority 
+// - Pool address
+// So that you can prove multiple notes 
+template SelectiveDisclosure(levels, nNotes) {
+    /** PUBLIC INPUTS **/
+    signal input roots[nNotes];
+    signal input noteCommitments[nNotes];
+    signal input extContextHash;
+ 
+    
+    /** PRIVATE INPUTS **/
+    signal input inAmount[nNotes];
+    signal input inPrivateKey[nNotes];
+    signal input inBlinding[nNotes];
+    signal input inPathIndices[nNotes];
+    signal input inPathElements[nNotes][levels];
+    
+    // Components
+    component inKeypair[nNotes];
+    component inCommitmentHasher[nNotes];
+    component inTree[nNotes];
+    
+    for (var ni = 0; ni < nNotes; ni++) {
+        // Verify that the sender actually owns the inputs
+        // He knows the secret keys and the blinding factors.
+        inKeypair[ni] = Keypair();
+        inKeypair[ni].privateKey <== inPrivateKey[ni];
+        
+        // Computes the leaf commitment as hash(amount, publicKey, blinding)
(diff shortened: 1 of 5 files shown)


maksimryndin:
@codex review the circuit

circuits/src/selectiveDisclosure.circom
```diff
@@ -0,0 +1,63 @@
...
+        inTree[ni].leaf <== inCommitmentHasher[ni].out;
+        inTree[ni].pathIndices <== inPathIndices[ni];
+        for (var i = 0; i < levels; i++) {
+            inTree[ni].pathElements[i] <== inPathElements[ni][i];
+        }
+        // Ensure root matches the expected value
+        roots[ni] === inTree[ni].root;
+    }
+    // Safety constraint to bind external context 
+    signal extContextSquare <== extContextHash * extContextHash;
```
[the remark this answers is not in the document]
Fantoni0: It's true that the proof can be replayed, with the same public inputs, as we don't implement any nullifier or additional replay prevention mechanism. But it's fine for selective disclosure, as it simply proves ownership and does not affect any fund management. 

The `extContextHash` will be checked to be correct on the smart contract (as we do with `pool.rs`). And it's bound to the Groth16 proof (it's part of the public inputs vector), so it cannot be modified without affecting the proof validity.

> the note owner can also generate fresh proofs for arbitrary contexts without any one-time binding

This is also intentional, as a user might need to prove the same note for different authorities.

circuits/src/selectiveDisclosure.circom
```diff
@@ -0,0 +1,63 @@
...
+    signal input noteCommitments[nNotes];
+    signal input extContextHash;
+ 
+    
+    /** PRIVATE INPUTS **/
+    signal input inAmount[nNotes];
+    signal input inPrivateKey[nNotes];
+    signal input inBlinding[nNotes];
+    signal input inPathIndices[nNotes];
+    signal input inPathElements[nNotes][levels];
```
NiDimi, line 28: Do these need to be private? Is this for efficiency?

circuits/src/selectiveDisclosure.circom
NiDimi, line 62: Perhaps comment here what goes into the hash when we know

NiDimi (approved):
Looks good, left some comments as we discussed to keep up the history

circuits/src/selectiveDisclosure.circom
Fantoni0: They do not to be private, as we are disclosing the commitment and the verifier can learn them. But it is more efficient, as we have fewer public inputs. This means:
- The verifier needs to perform one less pairing per public input.
- The transaction gets smaller, as it includes fewer public inputs.

circuits/src/selectiveDisclosure.circom
Fantoni0: Will probably add it in future PRs when the form of the external context it's settled

circuits/src/test/prove_selective_disclosure.rs
```diff
@@ -0,0 +1,226 @@
+#[cfg(test)]
+mod tests {
```
cupicmarko, line 2: Are these tests run on CI?

cupicmarko (approved):

circuits/src/test/prove_selective_disclosure.rs
Fantoni0: In the past we put heavy test on ignore to make it lighter on the CI. But after the latest changes, the CI also runs ignored tests. So yes, they are run

PR #205, opened 2026-05-20
Add notes table column sorting

cupicmarko:
Implement sortable columns (Note ID, Amount, Created, Status).

### DIFF
--- a/.github/workflows/ub-detection.yml
+++ b/.github/workflows/ub-detection.yml
@@ -25,6 +25,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+      - name: Generate circuit testdata
+        run: BUILD_TESTS=1 cargo build -p circuits
+
       - run: |
           rustup +nightly component add miri
           cargo +nightly miri setup
--- a/app/js/ui/notes-table.js
+++ b/app/js/ui/notes-table.js
@@ -24,11 +24,11 @@ export const NotesTable = {
                     b.classList.toggle('text-dark-50', isActive);
                     b.classList.toggle('text-dark-400', !isActive);
                 });
-                
+
                 this.render();
             });
         });
-        
+
         this.render();
 
         App.events.addEventListener('wallet:ready', () => {
@@ -95,7 +95,7 @@ export const NotesTable = {
             this._refreshing = false;
         }
     },
-    
+
     render() {
         const tbody = document.getElementById('notes-tbody');
         const empty = document.getElementById('empty-notes');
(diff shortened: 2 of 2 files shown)


maksimryndin (changes requested):
Why they don't come sorted from the backend https://github.com/NethermindEth/stellar-private-payments/blob/main/app/crates/core/state/src/storage.rs#L314 ?

We shouldn't change the frontend code to keep it slim. We should fix at the Rust level. Where do we lose the ordering?

cupicmarko:
> Why they don't come sorted from the backend https://github.com/NethermindEth/stellar-private-payments/blob/main/app/crates/core/state/src/storage.rs#L314 ?
> 
> We shouldn't change the frontend code to keep it slim. We should fix at the Rust level. Where do we lose the ordering?

There is actually nothing to update on backend side to the sorting, it is already queried by ledger. Frontend resorted the notes by leafIndex. Default sorting in the frontend now keeps it sorted as it was initially sorted by query.

maksimryndin (changes requested):
So my suggestion would be to avoid introducing the sorting on the js side as basically it will become soon obsolete as the amount of notes increases (we will introduce a pagination so the right way is to have ordering coupled with the pagination at the db level but the per column sorting is not needed now).

Let's make ensure
1) either the bug is reproduced (at least for me it happened once) and we can track the cause of it
2) or just close the PR without changes

I would prefer to figure out 1 :)

cupicmarko:
I didn't reproduce it, but the cause is sorting by [`notes.sort((a, b) => (b.leafIndex ?? 0) - (a.leafIndex ?? 0));`](https://github.com/NethermindEth/stellar-private-payments/pull/205/changes#diff-0ea9b52a354a6717f394ba3b3220fcf41a304faea900415b7cfadee2c91dbe7bL110). It could be simply completely removed and [the notes will stay sorted like they were by query](https://github.com/NethermindEth/stellar-private-payments/blob/main/app/crates/core/state/src/storage.rs#L314). 

I didn't reproduce it with about dozen notes, I assume notes can be placed in different position in Merkle tree and then when sorted by leafIndex ledgers appear out of order. I would suggest keeping sort buttons, users should be able to sort notes by status or amount.

maksimryndin:
> I didn't reproduce it, but the cause is sorting by [`notes.sort((a, b) => (b.leafIndex ?? 0) - (a.leafIndex ?? 0));`](https://github.com/NethermindEth/stellar-private-payments/pull/205/changes#diff-0ea9b52a354a6717f394ba3b3220fcf41a304faea900415b7cfadee2c91dbe7bL110). It could be simply completely removed and [the notes will stay sorted like they were by query](https://github.com/NethermindEth/stellar-private-payments/blob/main/app/crates/core/state/src/storage.rs#L314).
> 
> I didn't reproduce it with about dozen notes, I assume notes can be placed in different position in Merkle tree and then when sorted by leafIndex ledgers appear out of order. I would suggest keeping sort buttons, users should be able to sort notes by status or amount.

awesome! It is a remnant from my frontend vibe coding :) let's remove it to make the backend a single source of truth 👍

maksimryndin:
and in this PR let's also fix the UB flow https://github.com/NethermindEth/stellar-private-payments/actions/workflows/ub-detection.yml :)

cupicmarko:
Ok, should I revert all other changes and only remove the line that sorts it in the frontend?

maksimryndin (approved):
amazing! 👍

Issue #168, opened 2026-04-17
e2e for WebClient


maksimryndin:
we need to create a test (up to the tx signing and submission) for app/crates/platforms/web/src/client.rs

basically test main flows - deposit, transfer, withdraw (possible transact?)
we should use stellar-cli to deploy new contracts on the testnet (using scripts/deploy.sh)
then create 2 accounts 
add them to asp set etc
...

so it is a complete smoke test (except tx submission and signing - to be added later - see also https://github.com/NethermindEth/stellar-private-payments/issues/162)

this e2e test will be run before any new redeployment of the app

and check how to refactor https://github.com/NethermindEth/stellar-private-payments/blob/main/.github/workflows/wasm-build.yml#L41 - to encompass all wasm32 tests in the workspace

Fantoni0:
This can be closed right @cupicmarko

cupicmarko:
@Fantoni0 I would say it can, we have 14 e2e tests running in browser

PR #206, opened 2026-05-20
Multiple assets

maksimryndin:
- [x] note amount is now 16 bytes (instead of 8 bytes). The maximum `ExtAmount` to be covered by Stellar is `i128` (https://developers.stellar.org/docs/tokens/token-interface). So we use `u128` as a primitive for `NoteAmount` to cover the non-negative part of `i128` (and to allow adding amounts of 2 `ExtAmount::MAX`). Version of encryption is not added here (yet?). For privacy reason we are not going to support variable encodings. So when a new format comes, we should be able to re-encrypt unspent notes (based on the length) to a new version while being able to handle old notes for transitioning and to have then again fixed length for all notes in the system.
- [x] we have public keys registrations (address book) per pool. Probably, it is not needed now to have a single separate contract just for that registration or may be we should transfer it to ASP membership? (probably, we can address that later with UI changes, not here; intermediate step can be db UNIQUE for keys)
- [x] RPC reuquests are refactored to even more bulk requests (combining all the data keys) with a followup https://github.com/NethermindEth/stellar-private-payments/issues/227
- [x] indexer and storage logic for multiple pools
- [x] deployment config is moved from stellar crate to platform specific one as anyway later the sdk will provide a configurable deployment
- [x] ASP membership checks against specified contract address rather than all leaves in the db table
- [x] new contracts deployment

changes are mostly related to the backend

=== notes

- ASP contracts are global for all pools for now (to avoid introducing even more complexity)
- No radical UI changes for now (we have multiasset support on the backend but on the frontend side we just take the first pool from the array of pools and take the first pool address from the deployments.json - the UI changes will be introduced in a separate task of UI simplification) - I only added some small changes to make it compatible with the current UI
- is not tested with other tokens actually (it will be done separate with UI changes)
- web e2e test is still required https://github.com/NethermindEth/stellar-private-payments/issues/168 (when we stabilize)


maksimryndin:
@codex review

app/crates/core/stellar/src/indexer.rs
```diff
@@ -1,31 +1,42 @@
...
+
         // Retention-window check: if the RPC cannot serve events back to the deployment
         // ledger, onboarding on a fresh DB will fail to reconstruct Merkle
         // trees.
-        let contract_ids = [config.pool.to_string(), config.asp_membership.to_string()];
+        let contract_ids: Vec<String> = config
+            .pools
+            .iter()
+            .filter_map(|p| p.enabled.then_some(p.pool_contract_id.clone()))
+            .chain(std::iter::once(config.asp_membership.to_string()))
```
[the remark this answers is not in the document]
maksimryndin: https://github.com/NethermindEth/stellar-private-payments/issues/227

deployments/scripts/deploy.sh
```diff
@@ -37,17 +39,19 @@ Examples:
   deployments/scripts/deploy.sh futurenet \
     --deployer alice \
     --vk-file ceremony/verification_key.json \
-    --token CB... \
+    --pool native:CB... \
+    --pool contract:CC... \
+    --pool classic:USDC:G...:CD... \
```
Fantoni0, line 44: Maybe add a link to the Stellar documentation here. I found it confusing to differentiate between classic and native. And initially, I thought the contract was for trustlines.
Also, if you want to deploy multiple pools (one of them being native XLM) you still need to provide the contract address `--pool native:CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC.` That was not needed before, and I think that if there's no confusion possible it's simpler for the user NOT to provide the contracts. Or is it possible to have other native tokens than XLM?

app/js/ui/templates.js
```diff
@@ -152,12 +170,11 @@ export const Templates = {
             
             if (note) {
                 // Convert stroops to XLM for display
```
Fantoni0: There are still some mentions to XLM in code and in UI. I will list UI issues for completeness, but, as you mentioned in your PR description, it's better to address them in a separate PR. Feel free to ignore for now:

- Mentions to XLM in comments/code
- Some dead code not being used now in the UI. Some UI code calling removed functions.
- In the Withdraw tab, the text "Total to withdraw" still uses XLM.
- All note amounts in the different tabs (e.g., when you want to transfer a note) still use XLM.
- In the "Your notes" section, XLM still appears even if you deposit USDC.
- Right now, the UI does not provide the ability to change between different deployed pools. It also does not provide information about the asset used in the current pool. This can be confussing even for the demo
- If you change pools, you still need to clear cache + delete local storage. The UI does not detect the change and automatically updates. I think it should be possible, as I know the backend and SQL tables have the pool_id key.

app/crates/platforms/web/src/protocol.rs
```diff
@@ -85,10 +86,11 @@ pub enum ProverWorkerResponse {
...
     pub membership_blinding: Field,
-    pub amount_stroops: ExtAmount,
+    pub amount: ExtAmount,
     pub pool_root: Option<Field>,
     pub pool_address: Address,
     pub aspmem_root: Field,
+    pub aspmem_contract_id: Address,
     pub aspmem_ledger: u32,
     pub output_amounts: [NoteAmount; N_OUTPUTS],
     pub smt_depth: u32,
```
Fantoni0, line 96: If for now we are using a single membership ASP, and a single NON-membership ASP. Why do we manage `aspmem_contract_id`? Is it intentional? If needed, why don't we do the same for `aspnonmem_contract_id`. Related to how `storage.rs` manages the membership and non-membership proofs.

app/crates/core/state/src/storage.rs
```diff
@@ -778,233 +843,296 @@ impl Storage {
...
-                params![a.account_id],
-            )?;
+        let pool_count_u32 =
+            u32::try_from(pool_ids.len()).map_err(|_| anyhow::anyhow!("pool count exceeds u32"))?;
+        let base_quota = if pool_count_u32 == 0 {
+            0
+        } else {
+            total_limit
+                .checked_div(pool_count_u32)
+                .expect("pool count is not zero")
```
Fantoni0, line 875: Maybe adding a sanity check to ensure total_limit > pool_count to ensure all pools get a non-zero quota?

app/crates/core/prover/src/encryption.rs
```diff
@@ -246,23 +246,23 @@ pub fn decrypt_output_note(
...
-/// Encrypted data (112 bytes)
+/// Encrypted data (120 bytes)
 fn encrypt_note_data(recipient_pubkey_bytes: &[u8], plaintext: &[u8]) -> Result<Vec<u8>> {
     if recipient_pubkey_bytes.len() != 32 {
         return Err(anyhow!("Recipient public key must be 32 bytes"));
     }
-    if plaintext.len() != 40 {
+    if plaintext.len() != 48 {
         return Err(anyhow!(
             "Plaintext must be 40 bytes (8 amount + 32 blinding)"
```
Fantoni0:
```suggestion
            "Plaintext must be 48 bytes (16 amount + 32 blinding)"
```

Fantoni0 (approved):
Thanks @maksimryndin for the work 🙌 
I did test the new multi-pool features with XLM and USDC. The core of the protocol works 🥳 . But I found some issues and left some comments. Nonetheless, I think we can approve because the core works and does not break the demo.

Many of them can be ignored because, as you mention, they need to be addressed in another PR. But please take a look.

app/js/ui/templates.js
maksimryndin: thanks Antonio! Some in the core are addressed, on the js side I've intentionally left as is.

put your comment in the task https://github.com/NethermindEth/stellar-private-payments/issues/225

app/crates/core/prover/src/encryption.rs
maksimryndin: good catch! 👍 addressed https://github.com/NethermindEth/stellar-private-payments/pull/206/changes/a66977edc2230d1be018e74ce4ab00dbd11c1f1d

app/crates/core/state/src/storage.rs
maksimryndin: good catch! 👍 addressed https://github.com/NethermindEth/stellar-private-payments/pull/206/changes/a66977edc2230d1be018e74ce4ab00dbd11c1f1d

app/crates/platforms/web/src/protocol.rs
maksimryndin: good question! we need that field to pass to storage worker as we now filter the asp membership leaves by contract

deployments/scripts/deploy.sh
maksimryndin: put the comment into the task https://github.com/NethermindEth/stellar-private-payments/issues/225

as of now, the old command with a single token is supported (so basically from the deployer and a user perspective it stays the same).
When we introduce UI changes, I will change appropriately

PR #207, opened 2026-05-21
Selective disclosure crates

Fantoni0:
This PR introduces the core selective-disclosure support:
- Adds types for the JSON-receipt schema in the types crate.
- Adds a small disclosure crate that registers `selectiveDisclosure_1`, serializes public inputs in circuit order, and reuses the generic prover crate for real Groth16 proof generation and verification. (Adds a registry to support more selectiveDisclosure circuit instances).

The PR mocks some parts, and it's mostly theoretical for now (there is some dead code, and some functions/types will probably change). This is intentional as I wanted to make sure it can be merged into main without compromising the current demo functionality.  Next PRs will include breaking changes: new web workers for proving disclosure and UI changes to enable proving/verification in the web app directly.

### DIFF
--- a/app/crates/core/disclosure/src/lib.rs
+++ b/app/crates/core/disclosure/src/lib.rs
@@ -1,8 +1,4 @@
 //! Selective-disclosure circuit metadata and receipt validation.
-//!
-//! This crate intentionally does not perform proving or Groth16 verification
-//! yet. It owns the circuit-specific receipt checks that sit above the generic
-//! `prover` crate and below the platforms entry points.
 
 use anyhow::{Result, anyhow};
 use prover::prover::{Prover, verify_proof};
--- a/app/crates/core/disclosure/src/lib.rs
+++ b/app/crates/core/disclosure/src/lib.rs
@@ -140,6 +140,26 @@ pub fn validate_registered_receipt(
     Ok(circuit)
 }
 
+/// Mock-checks every root named by a disclosure receipt.
+///
+/// # Arguments
+/// * `receipt` - Receipt containing the roots to check.
+/// * `expected_vk_hash` - Verifying-key hash expected by the caller.
+///
+/// # Returns
+/// Returns `true` after receipt metadata validation.
+///
+/// # Errors
+/// Returns an error if receipt metadata is invalid.
+/// TODO: REMOVE THIS AFTER MODIFYING THE SMART CONTRACT.
+pub fn mock_receipt_roots_are_known(
+    receipt: &DisclosureReceipt,
+    expected_vk_hash: &str,
+) -> Result<bool> {
+    validate_registered_receipt(receipt, expected_vk_hash)?;
+    Ok(true)
+}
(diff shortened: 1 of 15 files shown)


app/crates/core/disclosure/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+/// * `expected_vk_hash` - Verifying-key hash expected by the caller.
+///
+/// # Returns
+/// Returns public inputs as 32-byte little-endian field elements, suitable for
+/// the generic Groth16 verifier.
+///
+/// # Errors
+/// Returns an error if the receipt does not match a registered circuit or if
+/// its public-input shape is invalid.
+pub fn receipt_public_inputs_bytes(
```
cupicmarko: This is a public function that looks like a cheap getter based on its name, but it actually validates the entire receipt, validates it, binds it to `expected_vk_hash` and then serializes it. If a caller already validated the receipt, it is a redundant work (I didn't check how expensive the validation is).

app/crates/core/disclosure/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+) -> Result<Vec<u8>> {
+    let circuit = validate_registered_receipt(receipt, expected_vk_hash)?;
+    let n_notes =
+        usize::try_from(circuit.n_notes).map_err(|_| anyhow!("Circuit n_notes out of range"))?;
+    let capacity = n_notes
+        .checked_mul(2)
+        .and_then(|n| n.checked_add(1))
+        .and_then(|n| n.checked_mul(32))
+        .ok_or_else(|| anyhow!("Public input byte capacity overflow"))?;
+    let mut out = Vec::with_capacity(capacity);
```
cupicmarko: I would suggest using `circuit.public_inputs_order` here if it is already defined.

cupicmarko (approved):
LGTM!

app/crates/core/disclosure/src/lib.rs
Fantoni0: You are right, will split the functionality to separate serialization from verification.

app/crates/core/disclosure/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+/// Returns `true` when the receipt proof verifies against `vk_bytes` and the
+/// receipt public inputs.
+///
+/// # Errors
+/// Returns an error if the receipt is malformed, targets an unsupported
+/// circuit, has unexpected metadata, or contains malformed proof bytes.
+pub fn verify_receipt_proof(
+    receipt: &DisclosureReceipt,
+    vk_bytes: &[u8],
+    expected_vk_hash: &str,
```
cupicmarko, line 340: I think this should be compared with hash of `vk_bytes`

cupicmarko (approved):

maksimryndin (approved):
lgtm! 👍

PR #214, opened 2026-05-23
feat: add push notifications to remind users after 5 days of inactivity

### LINKED ISSUE #160: Explore push notifications to remind a user to open the tab with the app if more than 5 days passed
related https://github.com/NethermindEth/stellar-private-payments/issues/169

Patrick-Ehimen:
Closes #160

## Summary
- Adds a service worker (`sw.js`) that fires a browser notification if the user has been away for more than 5 days
- Adds a `push-notifications.js` module handling permission requests, SW registration, and last-visit tracking via Cache API
- Wires a notifications step into the onboarding wizard to prompt users for permission on first use
- Calls `updateLastVisit()` and `registerServiceWorker()` on every page load

## How it works
On each visit the current timestamp is stored in the Cache API. A `periodicsync` event (`check-last-visit`, min interval 24 h) wakes the service worker in the background; if more than 5 days have elapsed and no app tab is visible, a desktop notification is shown.

## Test plan
- [x] Connect wallet — onboarding wizard shows a Notifications step and requests browser permission
- [x] Grant permission → service worker registers at `/sw.js` with scope `/`
- [x] In DevTools console, fake last visit: `const c = await caches.open('poolstellar-meta'); await c.put('/last-visit', new Response(String(Date.now() - 6*24*60*60*1000)))`
- [x] Navigate away from the tab, then in Application → Service Workers trigger periodic sync tag `check-last-visit` → notification fires

### DIFF
--- a/Trunk.toml
+++ b/Trunk.toml
@@ -44,12 +44,12 @@ if [ ! -d "$CIRCUIT_OUT" ]; then
     exit 1
 fi
 
-cp "$CIRCUIT_OUT"/*.r1cs $TRUNK_STAGING_DIR/circuits/
-cp "$CIRCUIT_OUT"/*.wasm $TRUNK_STAGING_DIR/circuits/
+cp "$CIRCUIT_OUT"/*.r1cs "$TRUNK_STAGING_DIR/circuits/"
+cp "$CIRCUIT_OUT"/*.wasm "$TRUNK_STAGING_DIR/circuits/"
 
 # Static files
-cp app/admin.html $TRUNK_STAGING_DIR/ 2>/dev/null || true
-cp app/disclosure.html $TRUNK_STAGING_DIR/ 2>/dev/null || true
+cp app/admin.html "$TRUNK_STAGING_DIR/" 2>/dev/null || true
+cp app/disclosure.html "$TRUNK_STAGING_DIR/" 2>/dev/null || true
 
 # Distribution legal files (licenses + notices).
 sh deployments/scripts/stage-dist-legal.sh "$TRUNK_STAGING_DIR"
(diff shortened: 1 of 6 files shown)


app/js/ui/push-notifications.js
```diff
@@ -0,0 +1,63 @@
...
+    } catch {
+        return false;
+    }
+}
+
+export function setNotificationsPrompted() {
+    try {
+        window.localStorage.setItem(NOTIFICATIONS_PROMPTED_FLAG, '1');
+    } catch {
+        // ignore
```
maksimryndin: at least let's output to console, otherwise we won't be able to see a problem

Trunk.toml
```diff
@@ -44,11 +44,11 @@ if [ ! -d "$CIRCUIT_OUT" ]; then
     exit 1
 fi
 
-cp "$CIRCUIT_OUT"/*.r1cs $TRUNK_STAGING_DIR/circuits/
-cp "$CIRCUIT_OUT"/*.wasm $TRUNK_STAGING_DIR/circuits/
+cp "$CIRCUIT_OUT"/*.r1cs "$TRUNK_STAGING_DIR/circuits/"
```
maksimryndin, line 47: why do we need changes here?

Trunk.toml
```diff
@@ -62,6 +62,9 @@ sh deployments/scripts/package-circuits-source-bundle.sh "$TRUNK_STAGING_DIR/cir
 stage = "build"
 command = "sh"
 command_arguments = ["-c", """
-./app/node_modules/.bin/esbuild app/js/ui.js --external:./web.js --bundle --outfile=$TRUNK_STAGING_DIR/js/ui.js --format=esm
-./app/node_modules/.bin/esbuild app/js/admin.js --external:./web.js --bundle --outfile=$TRUNK_STAGING_DIR/js/admin.js --format=esm
+./app/node_modules/.bin/esbuild app/js/ui.js --external:./web.js --bundle --outfile="$TRUNK_STAGING_DIR/js/ui.js" --format=esm
```
maksimryndin, line 66: why do we need these 2 lines changed?

maksimryndin (commented):
awesome @Patrick-Ehimen ! 👍 thanks for the contribution!

will test later, meanwhile some comments

maksimryndin:
@codex review

maksimryndin:
hey @Patrick-Ehimen ! 🙌  could you pls update us on any plans with this PR? (some comments to be addressed)

Patrick-Ehimen:
@maksimryndin sorry for the delay, i will look it into it ASAP

Trunk.toml
Patrick-Ehimen: The quoting was needed to handle paths with spaces in `TRUNK_STAGING_DIR` — without quotes the shell would split the path. This was a pre-existing bug unrelated to the push-notification work; I bundled the fix here since I was already touching the file. The same fix applies to the `cp` call on line 51.

Trunk.toml
Patrick-Ehimen: Two changes here: (1) the `outfile=` paths are now quoted for the same spaces-in-path reason above; (2) the new `cp app/js/sw.js` line is required — the service worker must be served from the root scope (`/sw.js`) and cannot be bundled into `ui.js` (bundling would break its scope and prevent it from intercepting page events). Without this line the service worker file simply wouldn't be present in the build output.

maksimryndin (approved):
awesome, thanks @Patrick-Ehimen !

PR #216, opened 2026-05-26
Update smart contract and getters

### LINKED ISSUE #208: Expose pool is-known-root method
We need the pool contract to expose whether a Merkle root is still accepted by the contract. This will be used by the selective disclosure verifier to determine whether a receipt was issued against a recent pool root, rather than just checking that the Groth16 proof is valid.

- Add pub fn is_known_root(env: Env, root: U256) -> Result<bool, Error>.
- It should wrap the existing MerkleTreeWithHistory::is_known_root logic.
- Add tests

Files to consider:
- `contracts/pool/src/pool.rs`
- `contracts/pool/src/test.rs`

Note: this is a breaking pool contract change, so it will require a redeploy of the contracts. Will affect the main demo. We need to postpone the merge until the whole milestone is implemented to prevent breaking the application for users.

### LINKED ISSUE #209: Add Stellar RPC helper for pool root read
Once the pool contract exposes is_known_root, the app needs a read-only way to call it through Stellar RPC. This is needed by selective disclosure verification: proof validity and context validity are not enough if the root is stale.

- Add `StateFetcher::is_pool_known_root(root: Field) -> Result<bool>.`
- Reuse the transaction building/simulation pattern already used for ASP non-membership proof fetching.
- Keep this as an RPC helper only. No storage changes or schema changes.

Files to consider:
- `app/crates/core/stellar/src/contract_state.rs`

Blocked by: #208

Fantoni0:
For Selective Disclosure:

- Close #208 as it exposes the is_known_root method on the smart contract.
- Close #209 as it updates `contract_state.rs`

Opened as a draft for now, as it includes breaking changes.

### DIFF
--- a/contracts/pool/src/pool.rs
+++ b/contracts/pool/src/pool.rs
@@ -706,6 +706,16 @@ impl PoolContract {
         Ok(MerkleTreeWithHistory::get_last_root(env)?)
     }
 
+    /// Check whether a pool Merkle root is still in the recent root history.
+    ///
+    /// # Arguments
+    ///
+    /// * `env` - The Soroban environment
+    /// * `root` - Pool Merkle root to check
+    pub fn is_known_root(env: &Env, root: U256) -> Result<bool, Error> {
+        Ok(MerkleTreeWithHistory::is_known_root(env, &root)?)
+    }
+
     /// Update the contract administrator
     ///
     /// Transfers administrative control to a new address. Requires
--- a/contracts/pool/src/pool.rs
+++ b/contracts/pool/src/pool.rs
@@ -712,8 +712,8 @@ impl PoolContract {
     ///
     /// * `env` - The Soroban environment
     /// * `root` - Pool Merkle root to check
-    pub fn is_known_root(env: &Env, root: U256) -> Result<bool, Error> {
-        Ok(MerkleTreeWithHistory::is_known_root(env, &root)?)
+    pub fn is_known_root(env: &Env, root: &U256) -> Result<bool, Error> {
+        Ok(MerkleTreeWithHistory::is_known_root(env, root)?)
     }
 
     /// Update the contract administrator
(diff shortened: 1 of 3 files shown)


contracts/pool/src/test.rs
```diff
@@ -268,24 +255,104 @@ fn merkle_insert_updates_root_and_index() {
...
+        MerkleTreeWithHistory::insert_two_leaves(
+            &env,
+            U256::from_u32(&env, 1),
+            U256::from_u32(&env, 2),
+        )
+        .unwrap_or_else(|err| panic!("expected first insertion to succeed: {err:?}"));
+    });
+    let evicted_root = pool.get_root();
+
+    for i in 0..90u32 {
```
cupicmarko, line 326: Nit. test only, but where does this `90u32` come from?

contracts/pool/src/pool.rs
```diff
@@ -706,6 +706,16 @@ impl PoolContract {
         Ok(MerkleTreeWithHistory::get_last_root(env)?)
     }
 
+    /// Check whether a pool Merkle root is still in the recent root history.
+    ///
+    /// # Arguments
+    ///
+    /// * `env` - The Soroban environment
+    /// * `root` - Pool Merkle root to check
+    pub fn is_known_root(env: &Env, root: U256) -> Result<bool, Error> {
```
cupicmarko: You can use &U256, it is already passed as a reference one line bellow.

cupicmarko (approved):

contracts/pool/src/test.rs
Fantoni0: If the amount of roots we store in MerkleTreeWithHistory (`const ROOT_HISTORY_SIZE: u32 = 90;`). I can create a getter method for it, but not sure if worth it for a unit test

PR #217, opened 2026-05-26
Fix UB detection CI failure

cupicmarko:
Replaced SystemTime::now() with an AtomicUsize counter in unique_test_dir to eliminate the clock_gettime syscall, and added a #[cfg(miri)] early-return guard to skip filesystem I/O when running under Miri isolation.

### DIFF
--- a/circuit-keys/src/lib.rs
+++ b/circuit-keys/src/lib.rs
@@ -271,19 +271,18 @@ mod tests {
     use std::{
         fs,
         path::{Path, PathBuf},
-        time::{SystemTime, UNIX_EPOCH},
+        sync::atomic::{AtomicUsize, Ordering},
     };
+
+    static COUNTER: AtomicUsize = AtomicUsize::new(0);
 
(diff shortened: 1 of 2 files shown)


circuit-keys/src/lib.rs
```diff
@@ -301,6 +300,8 @@ mod tests {
 
     #[test]
     fn atomic_write_replaces_contents_and_cleans_up_temp() -> Result<()> {
+        #[cfg(miri)]
+        return Ok(());
         let dir = unique_test_dir("atomic_write")?;
         let path = dir.join("vk.json");
```
Fantoni0, line 305: Would something like this also work? I think it's cleaner
```suggestion
#[test]
    #[cfg_attr(miri, ignore = "Filesystem I/O not supported under Miri isolation")]
    fn atomic_write_replaces_contents_and_cleans_up_temp() -> Result<()> {
```

Other than that, looks good to me

Fantoni0 (approved):

PR #230, opened 2026-05-27
Update CODEOWNERS

Fantoni0:
Since both @maksimryndin and @cupicmarko will be actively working on the project. I think it's a good idea to make them also code owners.


maksimryndin:
probably only Marko, I will jump in PRs on ad-hoc basis

maksimryndin (approved):

PR #231, opened 2026-05-27
Contributors

maksimryndin:
- [x] keep only a single PR template (as for our own prs we start blank) but for external contributors it helps more
- [x] add some community links, badges
- [x] removed screenshots as anyway we have a deployment


Fantoni0 (approved):
LGTM!

PR #237, opened 2026-05-31
Reject non-canonical BN256 nullifiers

1sgtpepper:
## Summary

Reject BN256 nullifiers that are not canonical field elements before they are passed to the verifier or written into pool state.

This adds a pool-level canonicality check for input nullifiers. Values greater than or equal to the BN256 scalar modulus now fail with `NonCanonicalPublicInput`.

## Changes

- Add `Error::NonCanonicalPublicInput`.
- Add a small helper to reject `U256 >= bn256_modulus`.
- Validate all input nullifiers before spent checks.
- Add a regression test for non-canonical nullifiers.

## Validation

Ran:

```sh
VERIFIER_VK_JSON=$PWD/deployments/testnet/circuit_keys/policy_tx_2_2_vk.json cargo test -p pool --locked -j 1 -- --nocapture
```

Result:

```text
9 passed; 0 failed
```

Also ran:

```sh
cargo fmt --package pool
git diff --check -- contracts/pool/src/pool.rs contracts/pool/src/test.rs
```

### DIFF
--- a/contracts/pool/src/pool.rs
+++ b/contracts/pool/src/pool.rs
@@ -51,6 +51,8 @@ pub enum Error {
     NotInitialized = 11,
     /// Arithmetic overflow occurred
     Overflow = 12,
+    /// Public input is not canonical in the BN254 scalar field
+    NonCanonicalPublicInput = 13,
 }
 
 /// Conversion from MerkleTreeWithHistory errors to pool contract errors
@@ -391,6 +393,37 @@ impl PoolContract {
         Ok(())
     }
 
+    /// Reject values outside the canonical BN254 scalar-field range.
+    ///
+    /// `Bn254Fr::from_bytes` expects field elements, so any `U256` that will be
+    /// converted into a verifier public input must be checked before
+    /// conversion.
+    fn validate_bn256_public_input(value: &U256, modulus: &U256) -> Result<(), Error> {
+        if value >= modulus {
+            return Err(Error::NonCanonicalPublicInput);
+        }
+
+        Ok(())
+    }
+
+    /// Validate every `U256` field that contributes to the verifier's public
+    /// input vector. The transaction path checks `ext_data_hash` against
+    /// `hash_ext_data` before proof verification, so this covers the remaining
+    /// public-input values.
+    fn validate_bn256_public_inputs(proof: &Proof, modulus: &U256) -> Result<(), Error> {
+        Self::validate_bn256_public_input(&proof.root, modulus)?;
+        Self::validate_bn256_public_input(&proof.public_amount, modulus)?;
+        for nullifier in proof.input_nullifiers.iter() {
+            Self::validate_bn256_public_input(&nullifier, modulus)?;
+        }
+        Self::validate_bn256_public_input(&proof.output_commitment0, modulus)?;
+        Self::validate_bn256_public_input(&proof.output_commitment1, modulus)?;
+        Self::validate_bn256_public_input(&proof.asp_membership_root, modulus)?;
+        Self::validate_bn256_public_input(&proof.asp_non_membership_root, modulus)?;
+
+        Ok(())
+    }
+
     /// Verify a zero-knowledge proof
     ///
     /// # Arguments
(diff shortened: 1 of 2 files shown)


contracts/pool/src/pool.rs
```diff
@@ -552,7 +562,9 @@ impl PoolContract {
             return Err(Error::UnknownRoot);
         }
         // 2. Nullifier checks (prevent double-spending)
+        let bn256_modulus = bn256_modulus(env);
         for n in proof.input_nullifiers.iter() {
+            Self::validate_bn256_public_input(&n, &bn256_modulus)?;
```
cupicmarko: This is good, but also every `U256` field should be checked before converting to `Bn254Fr`, something like:

```rust
Self::validate_bn256_public_input(&proof.root, &modulus)?;
// ...

for n in &proof.input_nullifiers {
    Self::validate_bn256_public_input(&n, &modulus)?;
}
```

contracts/pool/src/test.rs
cupicmarko, line 1: Tests for boundary values could be added too (`r`, `r-1`)

contracts/pool/src/pool.rs
```diff
@@ -51,6 +51,8 @@ pub enum Error {
     NotInitialized = 11,
     /// Arithmetic overflow occurred
     Overflow = 12,
+    /// Public input is not canonical in the BN256 scalar field
```
cupicmarko:
```suggestion
    /// Public input is not canonical in the BN254 scalar field
```

cupicmarko (commented):
PR correctly fixes non-canonical BN254 field elements before they reach the verifier, but it should be extended to validate all public inputs, not just nullifiers.

1sgtpepper:
Thanks for the review.

I’ve addressed the comments and added the requested boundary coverage. Could you take another look when you have a chance?

contracts/pool/src/test.rs
```diff
@@ -453,3 +453,157 @@ fn transact_rejects_bad_public_amount() {
...
+
+    let asp_membership_root = setup.asp_membership_client.get_root();
+    let asp_non_membership_root = setup.asp_non_membership_client.get_root();
+
+    let proof = Proof {
+        proof: mk_mock_groth16_proof(&env),
+        root,
+        input_nullifiers: {
+            let mut v: Vec<U256> = Vec::new(&env);
+            let canonical_nullifier = U256::from_u32(&env, 123456789);
```
maksimryndin: the test is named `transact_rejects_non_canonical_nullifier` but here the var name is `canonical_nullifier`, isn't it a contradiction?

contracts/pool/src/test.rs
```diff
@@ -453,3 +453,157 @@ fn transact_rejects_bad_public_amount() {
 
     assert!(pool.try_transact(&proof, &ext, &sender).is_err());
 }
+
+#[test]
+fn transact_rejects_non_canonical_nullifier() {
+    let env = test_env();
+    let setup = setup_test_contracts(&env);
+    let max = U256::from_u32(&env, 1000);
```
maksimryndin: I would have a more semantical name for this variable (and in the tests below) otherwise I don't understand its meaning in the env.register call below :)

maksimryndin (commented):
Hey @Kuhai9801 ! thanks a lot for the contribution! 👍  Some small nit picks

1sgtpepper:
Thanks @maksimryndin, I’ve addressed the naming comments.

maksimryndin (approved):
amazing 👍 thank you @Kuhai9801 !

maksimryndin:
@Kuhai9801 sorry,
"Commits must have verified signatures."
please refer to https://github.com/NethermindEth/stellar-private-payments/blob/main/CONTRIBUTING.md#commit-signing
you can squash your 5 commits into a single signed one

1sgtpepper:
Thanks, I squashed the branch into a single verified signed commit: ea6c9d7f774c913f79610cdcb9b3eecd2e75aaf0.

Issue #158, opened 2026-04-16
Add CLI interface


maksimryndin:
depends on https://github.com/NethermindEth/stellar-private-payments/pull/146

take a look at https://github.com/NethermindEth/stellar-private-payments/pull/138

related https://github.com/NethermindEth/stellar-private-payments/issues/162

mabsattar:
hi @maksimryndin  i would like to work on this, can you please assign me.

maksimryndin:
Hey @mabsattar ! The main branch is updated - feel free to start this task :)

mabsattar:
@maksimryndin noted, thank you for assigning me

maksimryndin:
closed with https://github.com/NethermindEth/stellar-private-payments/pull/334 and https://github.com/NethermindEth/stellar-private-payments/pull/351

PR #240, opened 2026-06-01
Move Soroban tx simulation and assembly into stellar core

### LINKED ISSUE #162: Consider moving tx simulation and submission into the app core
Atm as there is a single web platform, it is in JS

karar189:
## What this PR does

Moves Soroban transaction preparation from JavaScript into the platform-agnostic `stellar` core crate (closes #162).

After proving, the Rust layer now:

- Encodes pool contract arguments (`Proof`, `ExtData`, sender/account) in `soroban_encode.rs`
- Builds unsigned invoke-host-function envelopes
- Simulates via Soroban RPC (`simulateTransaction`)
- Merges simulation output (resource fees, `SorobanTransactionData`, auth entries) using a Rust port of the JS SDK `assembleTransaction` path in `tx_assemble.rs`

`WebClient` attaches the resulting `PreparedSorobanTx` (`txXdr` + `authEntries`) to deposit/withdraw/transfer prover output and exposes `prepareRegisterPublicKeys` for registration.

The JS UI no longer builds or simulates contract calls. It only signs auth entries and the transaction via Freighter, then submits and waits for confirmation through a slim `submitPreparedSorobanTx` helper.

Also adds WASM-compatible RPC helpers: `get_account`, `send_transaction`, and `get_transaction`.

---

## Related issues

Closes #162

---

## Context

**Why keep signing/submission in JS**

Freighter integration is browser-specific and already lives in `wallet.js`. The goal of #162 is to centralize platform-agnostic tx building and simulation in core so a future CLI (#158) can reuse the same path. Wallet signing and final RPC submit stay in the web UI for now.

**New modules in `app/crates/core/stellar/`**

| Module | Purpose |
|--------|---------|
| `soroban_encode.rs` | ScVal encoding for pool `Proof`, `ExtData`, and `Account` args |
| `tx_prepare.rs` | `prepare_pool_transact` and `prepare_register` on `StateFetcher` |
| `tx_assemble.rs` | Simulation response parsing and fee/auth merge (includes unit tests) |

**Web platform wiring**

`WebClient` calls `prepare_pool_soroban_tx` after the prover worker returns, emitting a "Simulating transaction…" progress stage. Prover protocol types (`DepositPrepared`, `PreparedProverTx`) now include a `sorobanTx` field.

**JS cleanup**

`stellar.js` shrinks from ~430 lines to ~110. Removed `contract.Client` construction, manual arg encoding, simulation, and assembly. Deposit/withdraw/transfer/register flows in `transactions.js` and `navigation.js` consume `proved.sorobanTx` from WASM instead of building txs locally.

**RPC additions**

`get_account` fetches sequence numbers for envelope construction. `send_transaction` and `get_transaction` are added to the core RPC client (submit/confirm still go through `@stellar/stellar-sdk` in JS for now; available for future platforms).

---

## Checklist

- [x] I've read the [contributing guide](../CONTRIBUTING.md)
- [x] I've tested locally and run `.githooks/pre-push`
- [x] I've updated or created tests if needed

---

## Test instructions for maintainers

```sh
# Enable pre-push hook (once)
git config core.hooksPath .githooks

# Run full local QA (fmt, clippy, typos, tests, contract builds)
git push  # or run .githooks/pre-push directly

# WASM-specific tests
cargo test --target wasm32-unknown-unknown -p stellar
cargo build -p web --target wasm32-unknown-unknown

# Manual smoke test
make install && make serve
```

Open http://localhost:8080, connect Freighter (testnet), then:

1. Register public keys (onboarding / navigation flow)
2. Deposit — sign auth + tx, confirm on-chain
3. Transfer or withdraw using the same sign/submit path

Expected: "Simulating transaction…" from WASM during prove; Freighter prompts only at sign/submit.


app/crates/core/stellar/src/soroban_encode.rs
cupicmarko, line 1: There should be more tests that encode real Proof, ExtData or account values, serialize them and verify byte-for-byte they're correct

app/crates/core/stellar/src/tx_prepare.rs
cupicmarko, line 1: This module is responsible for building every user-facing transaction but has no tests. I suggest adding tests that mock `Client::simulate_transaction` and asserts that returned result is correct.

app/crates/core/stellar/src/rpc.rs
```diff
@@ -578,6 +498,52 @@ impl Client {
...
+            LedgerEntryData::Account(entry) => Ok(entry),
+            _ => Err(Error::UnexpectedScVal(
+                "expected account ledger entry".into(),
+            )),
+        }
+    }
+
+    /// Submits a signed transaction envelope to the network.
+    #[allow(dead_code)]
+    pub async fn send_transaction(
```
cupicmarko, line 626: This should be wired into JS, or if it's going to happen later, marked with `TODO` comment and a GitHub issue.

app/crates/core/stellar/src/soroban_encode.rs
```diff
@@ -0,0 +1,149 @@
...
+        map_entry("c", bytes_to_scval(&proof_uncompressed[192..256])?)?,
+    ])
+}
+
+/// Encodes pool `Proof` public inputs + embedded proof for `transact`.
+#[allow(clippy::too_many_arguments)]
+pub fn pool_proof_to_scval(
+    proof_uncompressed: &[u8],
+    root: Field,
+    input_nullifiers: [Field; 2],
```
cupicmarko: Use `&[Field]` instead of hardcoded size.

app/js/stellar.js
```diff
@@ -328,98 +110,3 @@ export async function submitProvedPoolTransact(proved, ctx, opts = {}) {
 
     return hash;
```
cupicmarko: This will return hash if the loop times out, as if the transaction succeeded. It doesn't affect core functionality but UI might show incorrect state.

cupicmarko (changes requested):
The architecture is great and seems secure, but this is a critical path to be untested.

app/crates/core/stellar/src/soroban_encode.rs
karar189: Addressed in the latest commit. Added byte-for-byte XDR parity tests in soroban_encode.rs for Groth16Proof, Proof, ExtData, and Account each builds the contract SDK type, serializes with .to_xdr(), encodes with our off-chain helpers, and asserts the bytes match. Also added a test that our ExtData encoding matches the payload used by hash_ext_data_offchain. 

Happy to add cases with full-width field elements or fixed proof fixtures if you’d like more coverage beyond the current synthetic inputs.

app/crates/core/stellar/src/tx_prepare.rs
karar189: Added unit tests in` tx_prepare.rs` with a MockRpc that stubs `simulate_transaction`. 
They verify resource fees are merged into the envelope, and that register / transact produce the expected prepared XDR (fee, seq num, function name, arg count). No live RPC needed.

app/crates/core/stellar/src/rpc.rs
karar189: Not wired in this PR, submission and confirmation still go through `@stellar/stellar-sdk `in stellar.js (signing stays in the wallet). 
Also added TODO(#162) on `send_transaction `and `get_transaction` to track moving submit/poll into WASM in a follow-up.

app/crates/core/stellar/src/soroban_encode.rs
karar189: Done, switched to &[Field].

app/js/stellar.js
karar189: replaced the fallback return hash with a timeout error after 30 polling attempts.

app/crates/platforms/web/src/client.rs
```diff
@@ -1335,3 +1464,21 @@ fn parse_u32_decimal(s: &str) -> Result<u32, String> {
...
+        return Err(JsError::new(&format!(
+            "Invalid {what} (expected 0x + 64 hex chars)"
+        )));
+    }
+    let hex_body = &s[2..];
+    let mut out = [0u8; 32];
+    for (byte, chunk) in out.iter_mut().zip(hex_body.as_bytes().chunks(2)) {
+        let byte_str =
+            std::str::from_utf8(chunk).map_err(|_| JsError::new(&format!("Invalid {what}")))?;
+        *byte = u8::from_str_radix(byte_str, 16)
```
v0-e: Consider using here instead `types::parse_0x_hex_32` + `JsError` mapping.

karar189: Done @v0-e , `parse_hex32` now delegates to types::parse_0x_hex_32 with `JsError` mapping.

app/js/ui/transactions.js
```diff
@@ -8,18 +8,11 @@
...
 import { Templates } from './templates.js';
 
 const N_OUTPUTS = 2;
-let cachedContractConfig = null;
-
-async function getContractConfig() {
-    if (cachedContractConfig) return cachedContractConfig;
-    cachedContractConfig = await getHandle().webClient.contractConfig();
-    return cachedContractConfig;
-}
```
v0-e, line 22: `getContractConfig()` seems to be undefined now.

karar189: Fixed it , sorry, my bad, I removed it by mistake. Restored the `getContractConfig` helper.

v0-e:
@karar189 please test sending out some transactions using the app/testnet.
Please verify, I'm getting issues like this in the tx confirmation UI.
>Invalid Authorization Entry
>
>The authorization entry XDR could not be parsed.

v0-e:
I'll take it from here @karar189, thank you :+1:

v0-e:
Added a fix: replaces raw auth-entry XDR signing with the SDK’s authorizeEntry flow so Freighter signs the correct Soroban authorization preimage.

v0-e:
Refactored a bit the helpers also.
Squashed everything to sign the commits.

app/crates/core/stellar/src/tx_prepare.rs
```diff
@@ -0,0 +1,347 @@
...
+pub struct PoolTransactInput {
+    pub proof_uncompressed: Vec<u8>,
+    pub ext_data: ExtData,
+    pub public: OnchainProofPublicInputs,
+}
+
+impl StateFetcher {
+    /// Simulates `transact` and returns unsigned XDR + auth entries for the
+    /// wallet.
+    pub async fn prepare_pool_transact(
```
cupicmarko, line 25: Validate the contract ID against `self.config.pools`

app/crates/core/stellar/src/tx_assemble.rs
```diff
@@ -0,0 +1,308 @@
...
+        .map_err(|_| anyhow!("total fee does not fit into u32"))?;
+    tx.ext = xdr::TransactionExt::V1(soroban_data);
+
+    let op = tx.operations[0].clone();
+    let xdr::OperationBody::InvokeHostFunction(mut invoke) = op.body else {
+        return Err(anyhow!("expected invokeHostFunction operation"));
+    };
+
+    let existing_auth = invoke.auth.to_vec();
+    if existing_auth.is_empty() {
```
cupicmarko: I think that we should return an error if `existing_auth` is not empty, the create always builds envelopes with empty auth.

app/js/stellar.js
```diff
@@ -1,140 +1,84 @@
...
+
+    let validUntil = Number(entry.credentials().address().signatureExpirationLedger());
+    if (!validUntil) {
+        const ledgerSeq = latestLedger > 0
+            ? latestLedger
+            : (await server.getLatestLedger()).sequence;
+        validUntil = ledgerSeq + 100;
+    }
+
+    const signed = await authorizeEntry(
```
cupicmarko, line 40: `authorizeEntry` expects a callback that returns raw signature bytes but the code calls `signWalletAuthEntry` with preimage instead of full `SorobanAuthorizationEntry`

cupicmarko (changes requested):

app/crates/core/stellar/src/tx_assemble.rs
v0-e: Returns an error now :+1:

app/crates/core/stellar/src/tx_prepare.rs
v0-e: Add a common helper since that was being checked elsewhere :+1:

app/js/stellar.js
v0-e: Unsure about this, the naming is bit confusing indeed. See for example: https://github.com/stellar/js-stellar-sdk/pull/1044#issuecomment-2389024378
> signAuthEntry is poorly named, since what it actually does is sign a preimage of an auth entry

Update the fn doc comments to current behavior, let me know what you think.

maksimryndin (commented):
Awesome! thanks! 2 wishes/concerns
- ideally we should get rid off @stellar/stellar-sdk (js dependency) with moving to Rust
- we deeply take internal logic of tx handling - is it possible to re-use some crate from Stellar? Or least have references to the ported logic?

v0-e:
> ideally we should get rid off @stellar/stellar-sdk (js dependency) with moving to Rust

Yeah we need to that sooner or later, though we still need to it sign txs, unless you consider signing part of tx assembly its likely out of the scope of the issue/PR.

> we deeply take internal logic of tx handling - is it possible to re-use some crate from Stellar? Or least have references to the ported logic?

There is nothing official in Rust we could use here. In the community, there is https://github.com/rahul-soshte/rs-stellar-base that could help, if we are OK with it, seems to be being kept up-to-date.
I can suggest at least a refactor of `rpc.rs` using https://github.com/stellar/rs-stellar-rpc-client for the Rust client, and use imported types for both Rust/WASM clients if possible + actually start using the Rust side tx push (as @cupicmarko pointed out). But another PR is likely more suitable for that.

app/js/stellar.js
```diff
@@ -156,262 +100,47 @@ export async function submitPreparedSorobanTx(prepared, ctx, opts = {}) {
     if (!rpcUrl) throw new Error('Missing rpcUrl');
     if (!networkPassphrase) throw new Error('Missing networkPassphrase');
 
+    const walletAuthTotal = authEntries.filter((entryXdr) =>
+        needsWalletAuthEntry(xdr.SorobanAuthorizationEntry.fromXDR(entryXdr, 'base64'), address),
+    ).length;
+
+    const server = new rpc.Server(rpcUrl, { allowHttp: rpcUrl.startsWith('http://') });
```
maksimryndin: why do we need to allow for http here? (it is a remnant from the past but do we really need to allow for an unencrypted traffic?)

app/crates/core/stellar/src/tx_prepare.rs
```diff
@@ -0,0 +1,349 @@
...
+        rpc::{Error as RpcError, SimulateHostFunctionResult, SimulateTransactionResponse},
+        tx_assemble::test_fixtures::{empty_envelope, empty_soroban_data},
+    };
+    use futures::executor::block_on;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use stellar_strkey::ed25519;
+    use stellar_xdr::curr::{Limits, ReadXdr, WriteXdr};
+    use types::ContractConfig;
+
+    const DEPLOYMENT: &str = include_str!("../../../../../deployments/testnet/deployments.json");
```
maksimryndin: It doesn't make actual RPC calls to testnet here, right? It is just for an example config?

maksimryndin (dismissed):
lgtm! 👍 
some minor comments

maksimryndin (approved):

app/js/stellar.js
v0-e: Maybe it would be nice for local dev? Either way, I've removed it since this is for the web app :+1:

app/crates/core/stellar/src/tx_prepare.rs
v0-e: Should only hit the mock RPC. I've added a mock config also, so it doesn't raise confusion in this aspect.

app/js/stellar.js
maksimryndin: No need for now. Yeah, let's remove to avoid re-generating the same problem with the new UI :)

maksimryndin (dismissed):

cupicmarko (dismissed):

v0-e:
Resolved merge conflict.
Main change:
- replaced `build_invoke_contract_tx_envelope()` -> `simulate_transaction()` -> `sim.first_result()`
- by `build_find_key_simulation_tx()` -> `simulate_single_retval()`

maksimryndin (approved):

PR #245, opened 2026-06-05
Transaction planner

### LINKED ISSUE #232: Notes division and consolidation
note - depends on https://github.com/NethermindEth/stellar-private-payments/pull/206

# Problem

As a user I don't want to think about notes. I want available balances per token and ability to
- deposit(token, amount)
- withdraw(token, amount)
- transfer(token, amount, recipient)

The current circuit https://github.com/NethermindEth/stellar-private-payments/blob/main/circuits/src/policy_tx_2_2.circom assumes 2 inputs and 2 outputs. But it operates over notes as indivisible pieces. So e.g. when I want an arbitrary amount to transfer, I need first to get this amount via operations of division and/or consolidation of existing available notes.

Ideally, we want to hide that complexity from a user. Even more, ideally we want only *one* transaction signing (*explore the ways how it can be achieved*).

## Potential cases

- Transfer “whole note” still works: 10 → send 10, change 0, recipient gets 10.
- Partial transfer with change:
    - Start with one 10 note, send 3 → recipient gets 3, sender gets new 7 note.
- Two-input transfer:
    - Notes: 2 and 5, send 6 → recipient 6, sender change 1.
- Auto-selection:
    - Notes: 10, 4, 4; send 5 selects 10 (not 4+4) because it minimizes change? (*explore how could we minimize fees?* or select other reasonable optimization criteria)
- Auto-consolidation:
    - Notes: 1, 1, 1; send 3 triggers consolidation (1+1→2), then transfer (2+1→3).

## Assumptions

- Circuit limits remain N_INPUTS=2, N_OUTPUTS=2; “any amount” is achieved via extra consolidation.
- Transact advanced recipients unchanged.

## Acceptance
- we need this feature covered with tests (especially, the planning engine)
- the work probably can be done under a separate crate in app/core

v0-e:
Note merger feature that joins notes to fulfill the transfer/withdraw amount.
The current PR focus mostly on the planning rather than execution.

Prioritizes,
- doing less transactions;
- reduce the number of owned notes.

Assumes only the 2-in 2-out circuit currently.
For example,
- if the user has notes `1`, `1`, `2`, and needs to send `4`, two transactions are planned `1 + 1 -> 2 + 0` and then `2 + 2 -> 4 + 0`;
- if the user has notes `2`, `5`, and needs to send `4`, a transaction is planned `2 + 5 -> 4 + 3`. This reduces the number of owned notes by one, easing future withdraws.

I suggest making this crate actually part of the a new SDK crate.
Some nice abstractions are needed overall, thinking here mostly to support the execution of the `TransactionPlanner`. Execution requires creating transactions (potentially with pool syncs between each step) and pushing them.

A E2E test was introduced, based mostly on the existing one, though it would be good to refactor it in the future when better abstractions are in place.

Closes #232.

### DIFF
--- a/app/crates/core/tx-planner/src/plan/combination.rs
+++ b/app/crates/core/tx-planner/src/plan/combination.rs
@@ -39,17 +39,17 @@ pub fn find_combination(
     sorted_vals.sort_unstable_by_key(|&(_, x)| x);
 
     let two = TwoScan::scan(&sorted_vals, goal)?;
-    if let TwoScan::Exact(a, b) = two {
-        return Ok(CombinationResult::TwoExact(a, b));
+    if let TwoScan::Exact(i, j) = two {
+        return Ok(CombinationResult::TwoExact(i, j));
     }
 
-    let one = OneScan::scan(&sorted_vals, goal)?;
+    let one = OneScan::scan(&sorted_vals, goal);
     if let OneScan::Exact(i) = one {
         return Ok(CombinationResult::OneExact(i));
     }
 
-    if let TwoScan::Overshoot(a, b, excess) = two {
-        return Ok(CombinationResult::TwoOvershoot(a, b, excess));
+    if let TwoScan::Overshoot(i, j, excess) = two {
+        return Ok(CombinationResult::TwoOvershoot(i, j, excess));
     }
 
     if let OneScan::Overshoot(i, excess) = one {
@@ -72,15 +72,8 @@ enum TwoScan {
 impl TwoScan {
     /// Two-pointer scan over ascending `(index, amount)` pairs.
     fn scan(sorted: &[(usize, NoteAmount)], goal: NoteAmount) -> Result<Self, PlanError> {
-        Ok(Self::scan_impl(sorted, goal)?.unwrap_or(Self::Miss))
-    }
-
-    fn scan_impl(
-        sorted: &[(usize, NoteAmount)],
-        goal: NoteAmount,
-    ) -> Result<Option<Self>, PlanError> {
         if sorted.len() < 2 {
-            return Ok(None);
+            return Ok(Self::Miss);
         }
 
         let mut best_overshoot: Option<(usize, usize, NoteAmount)> = None;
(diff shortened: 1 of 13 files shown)


app/crates/core/tx-planner/src/plan/combination.rs
```diff
@@ -0,0 +1,338 @@
...
+    fn scan(sorted: &[(usize, NoteAmount)], goal: NoteAmount) -> Result<Self, PlanError> {
+        let mut exact = None;
+        let mut overshoot = None;
+
+        for &(idx, x) in sorted {
+            let Some(diff) = x.checked_sub(goal) else {
+                continue;
+            };
+            if diff.is_zero() {
+                exact = Some(idx);
```
maksimryndin: should we early return here as a priority of the exact match over overshoot?

app/crates/core/tx-planner/src/plan/combination.rs
```diff
@@ -0,0 +1,338 @@
...
+        }
+
+        if sum >= ctx.goal || index >= ctx.candidates.len() || *ctx.iteration_limit <= 0 {
+            return Ok(false);
+        }
+
+        *ctx.iteration_limit = ctx
+            .iteration_limit
+            .checked_sub(1)
+            .ok_or(PlanError::InputAmountOverflow)?;
```
maksimryndin: here probably another error variant? otherwise maybe misleading

app/crates/core/tx-planner/src/plan/combination.rs
```diff
@@ -0,0 +1,338 @@
...
+        let dataset = amounts(&[3, 6, 10]);
+        assert_eq!(find(&dataset, goal(10)), CombinationResult::OneExact(2));
+    }
+
+    #[test]
+    fn comb_two_overshoot() {
+        let dataset = amounts(&[1, 2, 5, 6, 12]);
+        assert_eq!(
+            find(&dataset, goal(10)),
+            CombinationResult::TwoOvershoot(2, 3, goal(1))
```
maksimryndin: `goal` here (in `CombinationResult` and the tests below) looks a bit misleading, probably, direct `NoteAmount::from` or if you prefer a helper, then `excess`?

app/crates/core/tx-planner/src/plan/mod.rs
```diff
@@ -0,0 +1,669 @@
...
+            }) => Some(Self { steps }),
+            _ => None,
+        }
+    }
+
+    /// Last step in the plan.
+    pub fn final_step(&self) -> &PlannedStep {
+        self.steps
+            .last()
+            .expect("transaction plan always has atleast one step")
```
maksimryndin: typo "at least"?

app/crates/core/tx-planner/src/plan/mod.rs
```diff
@@ -0,0 +1,669 @@
...
+        return Err(PlanError::NoSpendableNotes);
+    }
+
+    let values: Vec<NoteAmount> = notes.iter().map(|n| n.amount).collect();
+    let combo = find_combination(&values, amount)?;
+
+    let (indices, change) = match combo {
+        CombinationResult::Impossible => return Err(PlanError::NoCombination),
+        CombinationResult::OneExact(i) => (vec![i], None),
+        CombinationResult::TwoExact(a, b) => (vec![a, b], None),
```
maksimryndin: for clariy we can use `i` and `j` here and below as well as indices are assumed (to avoid any confusion)

maksimryndin (commented):
Overall I like it, despite being heuristic, it is readable and explainable, deterministic. That's good. 👍 

Some small comments and I would suggest to check `combination.rs` for proper early returns of `PlanError` (to ease future debugging on one side and to have semantically correct error variants on the other side)

maksimryndin:
@codex review please try to identify any blind spot not covered by the planner (combination.rs), any suboptimal paths

app/crates/core/tx-planner/src/plan/combination.rs
v0-e: Actually the following match would still prioritize exact match. Though of course it is confusing this way :+1: added early returns.

app/crates/core/tx-planner/src/plan/combination.rs
v0-e: Indeed :+1:

v0-e:
Lovely comments @maksimryndin. Fixed the error handling a bit so that internal `scan()` errors are carried out to outside.

maksimryndin (approved):
awesome @v0-e 🥇

PR #248, opened 2026-06-08
disclaimer

maksimryndin:
- [x] a disclaimer should reflect proper positioning


v0-e (approved):
A step to the proper positioning ✅

PR #250, opened 2026-06-09
increase timeout

maksimryndin:
- [x] increase the deployment timeout (https://github.com/NethermindEth/stellar-private-payments/actions/runs/27231988588) as probably now we have more crates


v0-e (approved):
➕

PR #251, opened 2026-06-10
Transaction plan execute

v0-e:
Adds tx plan execution. Related issue #232, follow up of PR #245. 

- `tx-planner` crate:
  - New `SpendSession` executor walks a `TransactionPlan` step-by-step;
  - Execution itself (signing, proving), is still outside logic. It would be nice to have a nicer integration.
- Web client:
  - Loads all unspent notes and runs transaction planning + execution on them;
  - Current approach is for each tx, prove-sign-push followed by a sync. "Safer" than sign all txs and then push as the wallet is always in sync, though the user will be prompted to sign across the execution plan. We can explore also adding the latter mechanism as an option;
  - If the indexer is slightly behind after a deposit, the app waits and retries instead of failing immediately.
  - Though if some reason the execution of the plan fails, there is no automatic re-trying or re-planning or any type of rollbacks. This should be OK as every planned tx contributes to a consolidation of different notes (good thing), OR the goal was reached;
  - Structurally we have also a bit less duplicated code: planned and manual flows are routed through `prove_transact_inner` (`proveWithdraw` / `proveTransfer` / `proveTransact` build a `Transact` step). And there was a split of main client.rs into mod.rs + transact.rs.

- **UI** has now:
  -  **Simple** mode (default): amount + recipient → planner picks notes, may confirm multi-tx;
  -  and a **Advanced** toggle: existing manual input/output note pickers.


app/crates/core/tx-planner/src/execute/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+enum SpendKind {
+    Transfer,
+    Withdraw,
+}
+
+/// Recipient data for the final step (and validation at `setup`).
+#[derive(Clone, Debug)]
+pub struct SpendTarget {
+    kind: SpendKind,
+    transfer_recipient_note: Option<NotePublicKey>,
```
maksimryndin: probably, looks like here it is better to have an enum with transfer and withdraw structs of corresponding data

app/crates/platforms/web/src/client/mod.rs
```diff
@@ -1202,6 +737,92 @@ impl WebClient {
...
+    ) -> Result<JsValue, JsError> {
+        let preview = self
+            .plan_spend_inner(pool_contract_id, user_address, amount)
+            .await?;
+        Ok(serde_wasm_bindgen::to_value(&preview)?)
+    }
+
+    #[wasm_bindgen(js_name = executeTransfer)]
+    #[allow(clippy::too_many_arguments)]
+    pub async fn execute_transfer(
```
maksimryndin, line 596: do we still need public `proveTransfer`? (and `proveWithdraw`, `proveDeposit`) if we have `execute...` methods

maksimryndin (commented):
awesome work @v0-e ! 👍 
That's great that you unified some flows!
The only thing I would wish - could we further simplify the exposed to js API surface of `WebClient`? E.g. having only `execute..` methods exposed without `prove...` methods?

app/crates/core/tx-planner/src/execute/mod.rs
v0-e: The correct approach ✅

app/crates/platforms/web/src/client/mod.rs
v0-e: Indeed, moved now `proveTransfer` and `proveWithdraw` to use `proveTransact` (manual, note picker approach). `execute...` use the tx planner, the `prove...` are manual Fns. Still more work to do here.
I didn't touch this part much yet since the focus was more on the tx plan execution (withdraw, transfer ops). Let me check this on a following PR.

v0-e:
Thanks for the fast review Maksim :)
On reducing the js API surface, let me check that on another PR, this one was more for multi-tx execution.

maksimryndin (approved):
👍 let's go!

PR #256, opened 2026-06-11
Fix advanced transfer submit rejecting balanced inputs

v0-e:
`updateTransferBalance` showed a green indicator in the UI but did not return `isBalanced`, so `runSpendFlow` always failed the submit-time check.


maksimryndin (approved):
fantastic :)

Issue #192, opened 2026-05-05
Witness generation alternatives


maksimryndin:
Atm we use [wasmer](https://github.com/wasmerio/wasmer) for both wasm and native compilation targets to generate witness (see https://github.com/NethermindEth/stellar-private-payments/blob/main/app/crates/core/witness/src/lib.rs#L52, related https://github.com/NethermindEth/stellar-private-payments/issues/154)

We use wasmer only for **wasm** as there is no yet a native interface (see https://github.com/NethermindEth/stellar-private-payments/issues/158, https://github.com/NethermindEth/stellar-private-payments/issues/159)

But wasmer in **native** has 3 available backends (they dropped support for the interpreter wasmi since 7.1.0)
- cranelift - slow (https://github.com/wasmerio/wasmer/tree/main/lib/compiler-cranelift#when-to-use-cranelift) but the main problem - conflict with soroban-sdk `arbitrary` crate dependency (https://github.com/stellar/rs-soroban-sdk/blob/main/soroban-sdk/Cargo.toml#L35)
- singlepass has a permissioned license 
- llvm - drags native deps and has a permissioned license https://github.com/wasmerio/wasmer/blob/main/lib/compiler-llvm/LICENSEhttps://github.com/wasmerio/wasmer/blob/main/lib/compiler-llvm/LICENSE

So we would like to explore the alternative https://github.com/philsippl/circom-witness-rs.
Note also that we have a freedom to change the wrapper `circom-compat` if needed (https://github.com/NethermindEth/circom-compat/tree/wasm-no-parallel) (ideally, get rid off wasmer and circom-compat)

1sgtpepper:
Hi, I’m looking at #192 and wanted to check the direction before going too deep.

Would it make sense to make generated witness graphs the main path, with Wasmer kept only as a temporary reference while equivalence is proven?

I saw the referenced Wasmer cleanup, but my read is that it mostly reduces dependency and fetch friction. The larger goal still seems to be avoiding Wasmer and `circom-compat` as the witness engine where possible.

A first step could be generating a `policy_tx_2_2.graph.bin`, wiring it into witness generation, and testing the output against the current Wasmer path.

Does that sound like the direction you’d prefer?

maksimryndin:
Hey @Kuhai9801 ! Yes, it seems so. thanks! Please note to make your PRs reviewable, please follow the PR template

PR #258, opened 2026-06-12
Add native policy witness graph runtime

1sgtpepper:
## What's this PR do?

Replaces `policy_tx_2_2` witness generation with a generated `circom-witness-rs` graph runtime for native and browser/WASM use.

Native and web builds now load the checked-in `policy_tx_2_2.graph.bin` artifact with the matching R1CS shape. Browser packaging stages `.r1cs`, `.graph.bin`, and `.graph.manifest` instead of the previous witness WASM artifact.

The graph manifest now pins the graph hash, Circom version, graph-builder lock hash, circuit source hashes, `circomlib.lock`, `circuits/build.rs`, and the graph-generation script. The circuits build verifies the manifest before publishing artifacts.

Wasmer remains only in dev/test scope for the equivalence and benchmark path.

## Context

This follows #192 and the review discussion on this PR: use generated witness graphs as the main policy witness path, cover browser/WASM, and prove whether the performance tradeoff is worth it.

This patch includes:

- native and browser-compatible witness calculation through the checked-in policy graph
- native black-box function bindings needed by the generated graph
- deterministic policy bus input normalization:
  - coalesces flattened policy JSON into grouped graph bus inputs
  - rejects conflicting grouped/flattened representations
  - preserves graph-required aliases only when equivalent
- graph generation tooling pinned to Circom `2.2.3`
- manifest validation for graph hash, source hashes, Circom version, and builder lock hash
- browser artifact staging for `.r1cs`, `.graph.bin`, and `.graph.manifest`
- a PR smoke workflow that computes a native graph witness, proves the reference policy case, and verifies it
- a reproducible benchmark comparing Wasmer+Cranelift against the graph runtime
- deployment key documentation for the graph and manifest artifacts

## Evidence

Latest smoke run:

https://github.com/Kuhai9801/stellar-private-payments/actions/runs/27586793963

- `native_policy_graph_witness_proves_reference_case ... ok`

Latest benchmark run:

https://github.com/Kuhai9801/stellar-private-payments/actions/runs/27587049051

The benchmark first checks byte-equivalent witness output against Wasmer, then reports timings:

- witness size: 37,679 field elements / 1,205,728 bytes
- init median: 203.387 ms Wasmer+Cranelift vs 2.875 ms graph, about 71x faster
- reused witness median: 186.689 ms Wasmer+Cranelift vs 3.655 ms graph, about 51x faster
- cold init+witness median: 393.949 ms Wasmer+Cranelift vs 6.624 ms graph, about 59x faster

Local validation:

- `cargo test -p witness`
- `cargo fmt --check`
- `git diff --check`

Termux note: full local `circuits` builds are blocked on this host by a `rustc`/Cranelift stack overflow while compiling the Wasmer benchmark dependency, so the heavy proof/equivalence/performance checks are covered by GitHub Actions above.

## Related issues

Related to #192

## Required Checklist

- [x] I've read the [contributing guide](../CONTRIBUTING.md).
- [x] I've mentioned this PR in the project LinkedIn group.
- [ ] I've tested this locally, run .githooks/pre-push and provided test instructions for maintainers if needed
- [x] I've updated or created tests if needed


maksimryndin:
hey @Kuhai9801 ! 🙌  fantastic work! Thank you! I wonder have you checked
1) is it possible to get rid off circom compat and wasmer in **wasm/browser** environment?
2) if we compare wasmer with cranelift and circom-witness-rs natively, what is the performance gain?

1sgtpepper:
Thanks, I checked both points.

1. For wasm/browser, I looked into whether this PR should also remove `circom-compat`/Wasmer. I do not think that should be bundled into this PR because this change is scoped to the native graph runtime, while browser still uses the existing wasm-compatible witness path.

This PR follows the direction from #192 by moving native witness generation to generated graphs first and keeping browser behavior stable. A follow-up can target browser graph execution and remove browser `circom-compat`/Wasmer separately.

2. For native performance, I added a GitHub Actions benchmark comparing Wasmer+Cranelift against `circom-witness-rs` graph execution. The benchmark checks byte-equivalent witness output before timing.

Results:
- init median: 201.675 ms Wasmer+Cranelift vs 2.844 ms graph, ~70.9x faster
- reused witness median: 104.108 ms vs 3.097 ms, ~33.6x faster
- cold init+witness median: 306.467 ms vs 5.963 ms, ~51.4x faster

Run: https://github.com/Kuhai9801/stellar-private-payments/actions/runs/27424454691

Would you be open to follow-up PRs for:
- browser graph backend: use generated witness graphs for wasm/browser as well, with the goal of removing browser Wasmer/`circom-compat`;
- native typed input path: avoid JSON/string/BigInt conversion in the native hot path by using a prepacked circuit input layout, with byte-equivalence checks against the current path.

maksimryndin (changes requested):
Let's implement the full solution (including the browser environment)? We need to see the whole picture to evaluate either it is worth or not

Izzzaaaa:
Thanks, agreed. I’ll extend this PR to cover the browser/WASM path too so the full replacement can be evaluated.

I’ll update the web prover to use the generated graph artifact, try to remove the browser `wasmer`/`ark-circom` path, and add validation for the existing policy proof flow. If `circom-witness-rs` has a wasm blocker, I’ll post the exact repro and we can decide whether to keep this PR native-only or patch that dependency first.

1sgtpepper:
Apologies, I accidentally commented using my other account I use for my independent project. Working on it!

1sgtpepper:
Ready for another review pass.

Bus index validation is the remaining hardening point. `nonMembershipProofs[99][0].*` can currently be flattened into grouped `nonMembershipProofs` if the final width matches. This does not look like a proof soundness issue, but it can hide caller bugs.

If `compute_witness` is public/semi-public, we should add dense zero-based index validation. If the JSON shape is internal/generated-only, this can be follow-up hardening.

maksimryndin:
hey @Kuhai9801! ✋   thanks a lot for the contribution, we need some time to review it.

1sgtpepper:
Thanks! take your time

maksimryndin:
closed in favor of https://github.com/NethermindEth/stellar-private-payments/pull/323

1sgtpepper:
I think #258 should land first, then #323 can rebase on top. #258 is the smaller reviewable step because it introduces the witness-graph runtime and provenance checks without also changing circuit keys or deployment state. Replacing it with #323 mixes runtime replacement with key and deployment changes, which will make review and regression attribution extremely messy.

PR #260, opened 2026-06-12
blinding derivation

maksimryndin:
- [x] membership blinding now is derived from the signature as well
- [x] we don't expose private keys to the frontend anymore

==
don't blame frontend :) we will refactor it later completely
the change is breaking (at least we don't introduce migrations here so the local storage should be cleared)


v0-e (approved):
:100:% perfect!

Issue #169, opened 2026-04-17
bootnode


maksimryndin:
### User Story

As a user, I don't want to care about the RPC retention window. I would like to be able to join the app anytime, or loose my local data and being able to restore it.

### Description

Probably, a solution could be a simple bootnode which allows a user an extended (since the contract deployment ledger) RPC service for a narrow set of RPC methods (actually used in the app indexer). The app on the RpcSyncGap error should offer a user an option with a clearly stated trust assumptions (see below) to use a bootnode.

### Acceptance Criteria

- the bootnode should expose the necessary JSON RPC methods (`getEvents`  and `getLatestLedger`) - for any other methods is should respond unsupported - it should be fully compatible with the standard RPC in its requests and responses formats for the mentioned rpc methods
- the logic: the app starts (after user's consent) calling the bootnode provides RPC for the app up until 5 days before the latest network tip, then it starts return HTTP redirect response to the true RPC (it is configured with). Here we should also resolve how we align the bootnode rpc with the one the has in the wallet (probably it is not a problem as we separate an indexer and the tx submission clients in the code). 5 days are within the retention window, it shouldn't be a problem for the app to catch up then. **We should carefully plan cursors/paging switch here to avoid any gaps.**
- the tools/bootnode crate should be a part of the Rust workspace and **probably** depend on `app/crates/core/stellar` crate and definitely statically include contract ids (from DEPLOYMENT `app/crates/core/stellar/src/lib.rs`) to prevent any misconfiguration.
- dynamic configuration of the server can include host:port, rpc address, acme cachedir, rate limiting params etc
- the server should only serve https (use [rustls-acme](https://crates.io/crates/rustls-acme))
- the server should expose CORS header with all origins allowed
- because it is a public service - IP rate limiting middleware should be added (prefer Axum as a web server) and secure headers (e.g. https://owasp.org/www-project-secure-headers/)
- the indexing logic can be simple - just looping over the rpc in a similar fashion app/crates/core/stellar/src/indexer.rs for the configured contracts. But don't process events. Just store raw responses from RPC.
- metrics are calculated (consider `metrics-rs`) - how many requests are in the moment, histogram of requests handling durations, last ledger indexed?, requests/responses bytes sizes
- the node Rust binary should live in `tools` directory
- the server should expose standard probe endpoint `healthcheck` (HTTP 200 OK if all good)  and `metrics` endpoint (with Prometheus compatible metrics)
- open telemetry 
[...]

PR #273, opened 2026-06-16
Bootnode

v0-e:
Bootnode provides older history that the RPC cannot provide (`getEvents` requests). If the client requests recent history, the bootnode responds with a handoff message, triggering the client to connect to the RPC.

`bootnode` crate is separate from workspace, potentially could be moved to its own repo.

Contributes for #169.

Follow ups:
- support multiple contracts;
- support bootnode for bootnode;
- improve shutdown handling;
- integration tests (actual client and bootnode, and (mock?) RPC;
- optional database pruning of older data;
- metrics.


app/js/ui/navigation.js
```diff
@@ -10,6 +10,120 @@ import { App, Utils, Toast } from './core.js';
 import { setTabsRef } from './templates.js';
 import { runOnboardingWizard } from './onboarding-wizard.js';
 
+const BOOTNODE_ENABLED_KEY = 'poolstellar_bootnode_enabled';
+const BOOTNODE_URL_KEY = 'poolstellar_bootnode_url';
+const DEFAULT_BOOTNODE_URL_TESTNET = 'https://bootnode.testnet.poolstellar.org';
+
+function getBootnodeSettings() {
```
maksimryndin: we shouldn't store the bootnode with the indexdb storage - we should store it in SQLite (like e.g. disclaimer consents) - the reason is that if we would like user states between web and cli versions, the configuration is preserved

tools/bootnode/src/rpc.rs
```diff
@@ -0,0 +1,212 @@
...
+};
+use metrics::counter;
+use serde_json::json;
+use std::sync::atomic::Ordering;
+use stellar::{
+    ContractEventFilter, GetEventsParams, GetEventsResponse, GetLatestLedgerResponse,
+    PaginationParams,
+};
+
+pub const CACHE_MISS_CODE: i32 = -32_004;
```
maksimryndin, line 21: why such strange codes? May it is worth to comment?

tools/bootnode/deny.toml
```diff
@@ -0,0 +1,80 @@
+# Bootnode dependency policy
```
maksimryndin: actually, we can just provide cargo deny with a path to the root config in the bootnode CI to avoid having it here again

tools/ceremony-cli/Cargo.toml
```diff
@@ -17,7 +17,7 @@ ark-circom.workspace = true
 ark-groth16.workspace = true
 ark-serialize.workspace = true
 circuit-keys.workspace = true
-clap = { version = "4.6.1", features = ["derive"] }
+clap = { version = "4.6.1", features = ["derive", "env"] }
```
maksimryndin: why do we need this change here?

app/js/ui/navigation.js
```diff
@@ -10,6 +10,120 @@ import { App, Utils, Toast } from './core.js';
...
+}
+
+function showBootnodeConsentModal({ defaultUrl, rpcUrl, errorMessage }) {
+    const existing = document.getElementById('bootnode-consent-modal');
+    if (existing) existing.remove();
+
+    const overlay = document.createElement('div');
+    overlay.id = 'bootnode-consent-modal';
+    overlay.className = 'fixed inset-0 z-50';
+    overlay.innerHTML = `
```
maksimryndin: innerHTML :) let's move it to the html as the piece is static

app/crates/core/stellar/src/indexer.rs
```diff
@@ -1,66 +1,92 @@
...
 use anyhow::{Result, anyhow};
-use std::collections::HashSet;
+use std::{cell::Cell, collections::HashSet};
 use types::{ContractConfig, ContractEvent, ContractsEventData, SyncMetadata};
 
 // https://developers.stellar.org/docs/data/apis/rpc/api-reference/methods/getEvents
 const PAGE_SIZE: usize = 300;
 const MAX_PAGES_PER_ROUND: usize = 10;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
maksimryndin: I expected no changes here, tbh :)
We just provide bootnode url as an rpc url and that's it. So indexer doesn't know about the bootnode existence. For an indexer and the rest of the app there is no boot node. Only some rpc service

The calling side can swap the bootnode with RPC

maksimryndin (commented):
Awesome work @v0-e on the bootnode! 🚀 

On its integration to the app though there are questions

tools/ceremony-cli/Cargo.toml
v0-e: Some workspace refactor remains, removed it :+1:

tools/bootnode/deny.toml
v0-e: Yeah, had it like that, though I figured with bootnode workspace isolation I add one specific for it. Using main deny.toml again :+1:

tools/bootnode/src/rpc.rs
v0-e: JSON-RPC codes we are using to notify the client, added docs :+1:

app/js/ui/navigation.js
v0-e: I see, moved it to SQLite ▶️

app/js/ui/navigation.js
v0-e: Done ✅

app/crates/core/stellar/src/indexer.rs
v0-e: I see, moved bootnode connection logic outside of stellar crate :+1:
Let me know if you have in mind a better place 🙏

app/crates/core/stellar/src/indexer.rs
```diff
@@ -1,4 +1,4 @@
-use crate::rpc::{Client, Error as RpcError};
+use crate::rpc::Client;
```
maksimryndin: why do we need any changes in the indexer at all? (or at least so many changes?) If the bootnode preservers requests and responses?

app/crates/core/state/src/storage.rs
```diff
@@ -299,6 +305,44 @@ impl Storage {
...
+            url: String::new(),
+        }))
+    }
+
+    pub fn set_bootnode_config(&mut self, enabled: bool, url: &str) -> Result<()> {
+        let tx = self
+            .conn
+            .transaction()
+            .context("failed to start transaction")?;
+        tx.execute("DELETE FROM bootnode_config", [])
```
maksimryndin: if you need to change a single row, you can add a limit constraint on the table (LIMIT 1) and then update the row here to have a cleaner query logic?

app/crates/core/stellar/src/rpc.rs
```diff
@@ -14,14 +14,22 @@ use stellar_xdr::curr::{
     LedgerKey, LedgerKeyAccount, Limits, PublicKey, ReadXdr, Uint256, WriteXdr,
 };
 
+/// `-32002`: bootnode JSON-RPC code telling the client to resume on its wallet
```
maksimryndin: let's avoid changing this crate if we plan to integrate official one. Also we change the request-response formats. We can signal the handoff with HTTP codes (ok, not a redirect but another one).

maksimryndin (commented):
the integration of the bootnode into the core is still coupled, let's try to make it even lighter and simpler? (for the core it should be basically one RPC switched for another one. The core crates shouldn't differentiate between a bootnode and an RPC)

app/crates/core/stellar/src/indexer.rs
v0-e: At least we need to change the `Indexer::init()`, since that always checks for the contract-configured [min_pool_ledger](https://github.com/NethermindEth/stellar-private-payments/blob/main/app/crates/core/stellar/src/indexer.rs#L39). I changed so that the stored ledger (which is updated as we get events during syncs) is read and used for this.

app/crates/core/state/src/storage.rs
v0-e: Sure, moved to a singleton approach :+1:

app/crates/core/stellar/src/rpc.rs
v0-e: Alright, no changes to `rpc.rs` :+1:
I think the request-response formats were coherent with Stellar's. Just added some types to replace the manual JSON ser/de.
Perhaps the JSON-RPC code is a better approach since we are dealing with JSON-RPC, it is more "app-level". Let me know though if you prefer HTTP-based handoff.

maksimryndin (dismissed):
awesome @v0-e ! 🥇 :)

app/crates/core/stellar/src/rpc.rs
maksimryndin: I don't have a strong preference here. The main requirement which you fulfilled was a minimum integration coupling of the bootnode with the core. At the sdk level we then will decide how to abstract that properly

tools/bootnode/src/storage/postgres.rs
```diff
@@ -0,0 +1,231 @@
...
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use deadpool_postgres::Pool;
+use tokio_postgres::types::Json;
+
+pub struct Postgres {
+    pool: Pool,
+}
+
+impl Postgres {
```
cupicmarko, line 12: There should be some retention policy this might grow indefinitely

tools/bootnode/src/storage/in_memory.rs
```diff
@@ -0,0 +1,235 @@
...
+struct State {
+    last_cursor: Option<String>,
+    last_fully_indexed_ledger: u32,
+    ledger_tip: u32,
+    cache_by_cursor_in: HashMap<String, GetEventsResponse>,
+    cache_by_start_ledger: HashMap<u32, GetEventsResponse>,
+    ledger_by_cursor_out: HashMap<String, u32>,
+}
+
+pub struct InMemory {
```
cupicmarko, line 18: Similar to [other comment in postgress](https://github.com/NethermindEth/stellar-private-payments/pull/273/changes#diff-e70177cad6a64f0740aa93aa10ad7b8fb83a76ca25ce0cdc41f43a259df7a8b3R12l), which can exhaust disk, this here will much faster exhaust RAM.

maksimryndin: I don't think we will use in-memory storage in production :) I would even go simpler having just db-backed storage (even with a concrete type) but @v0-e  likes storage traits :)

tools/bootnode/src/storage/postgres.rs
v0-e: Yeah, I was thinking about adding it later :+1:
Though how useful would that be, would it make sense for a bootnode to just provide partial data for a certain pool?
I'm guessing we need the complete pool state to issue transactions? cc @Fantoni0

tools/bootnode/src/storage/in_memory.rs
v0-e: Yeah it is just for tests :+1:
I'm also not a fan of it, it was more to support Rust unit tests without having to deploy a Postgres instance.
I was thinking of testing with Postgres at least in CI, but locally we can also:
- use `testcontainers`, which requires Docker installed and running;
- switch to SQLite 😄
Let me know what you guys prefer 🙏

maksimryndin: Let's not spend too much time on it as we have a solution (in memory for testing). As of now, all good.
We can explicitly put a `#[cfg(test)]` around in memory implementation and that's it.

cupicmarko: For testing completely fine :slightly_smiling_face:

v0-e:
Fixing an issue where the bootnode with respond with a cache miss when an empty page near cutoff was requested by the client, even though the bootnode was in sync with the rpc. (we are not storing empty pages)

Also removed the default bootnode URL in the UI, since we don't have any instance deployed yet.

maksimryndin (approved):

PR #275, opened 2026-06-17
CI unused deps check

v0-e:
Checks for unused dependencies using [cargo shear](https://github.com/Boshen/cargo-shear). Seems to be a bit better than machete, [see the claims](https://github.com/Boshen/cargo-shear#prior-art).
This PR also removes the current unused deps.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -10,6 +10,12 @@ if ! cargo sort --workspace --check; then
     exit 1
 fi
 
+# Unused dependencies
+if ! cargo shear; then
+    echo "❌ Unused dependencies detected (run 'cargo shear --fix')"
+    exit 7
+fi
+
 # Dependency audit
 if ! cargo deny check; then
     echo "❌ Critical: Vulnerable dependencies detected (run 'cargo deny check')"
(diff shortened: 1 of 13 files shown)


Fantoni0 (commented):
Should we also update the pre-push script to run cargo shear?

v0-e:
We can leave it in the CI only for now, since it requires another extra thing that the user needs to install. Plus, currently pre-push checks and CI checks are not the same broadly.
Though I can add it there if you like :+1:

Fantoni0:
> We can leave it in the CI only for now, since it requires another extra thing that the user needs to install. Plus, currently pre-push checks and CI checks are not the same broadly. Though I can add it there if you like 👍

I prefer to have it in the script, but up to you :+1: . The most similar pre-push and CI checks are, I think, the best. After installing typos and deny I think shear it's fine. And at the end, the user might still ignore the pre-push script  and create the PR directly :laughing:

Fantoni0 (dismissed):
LGTM!

v0-e:
Added it to pre-push :)

Fantoni0 (approved):
Thanks @v0-e !

PR #276, opened 2026-06-17
address registry

### LINKED ISSUE #271: Unlink address book from a pool contract as we have now multiple assets (either as a separate contract or move to an ASP)

maksimryndin:
- [x] pubkeys registration is moved into a separate contract as we have multiple pools now (and semantically it is better to separate)

closes https://github.com/NethermindEth/stellar-private-payments/issues/271

note - the new deployment is needed to have the demo working properly so it is included. Regular storage reset will be needed to ensure we don't have previous remnants

UPD
following @Fantoni0 right catch on duplication - contract storage is introduced to check for any duplications. Nevertheless, events based approach is preserved for consistency with other historical streams. Also it allows for broadcasting


contracts/public-key-registry/src/test.rs
```diff
@@ -0,0 +1,18 @@
...
+    let owner = Address::generate(&env);
+    let account = Account {
+        owner,
+        encryption_key: Bytes::from_array(&env, &[0x11; 32]),
+        note_key: Bytes::from_array(&env, &[0x22; 32]),
+    };
+
+    env.mock_all_auths();
+    client.register(&account);
+}
```
Fantoni0, line 116: Given the name of the test would it make more sense to not mock the auth and expect it to fail? Maybe I am confused by the test name

contracts/public-key-registry/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+        account.owner.require_auth();
+        PublicKeyEvent {
+            owner: account.owner,
+            encryption_key: account.encryption_key,
+            note_key: account.note_key,
+        }
+        .publish(&env);
+    }
+}
+
```
Fantoni0, line 91: Nothing prevents me from calling this multiple times. So we need to control in the UI when indexing events, so that we don't show the same owner multiple times. By switching accounts in Freighter, I was able to appear multiple times in the address book. 
I think we make the check for our own address, but not for others. So when you switch accounts you can see others appear multiple times

Image

Image

Fantoni0 (changes requested):
Great idea, and it mostly works. Just a couple of nitpicks and something I think needs to be handled: Multiple repeated address book lines can appear. I think this was actually already happening before creating the address registry.

Also, a couple of stale mentions to pool events in  `events_parsers.rs:14` and `soroban_encode.rs` (test `pool_account_encoding_matches_contracttype_xdr`). These were flagged by AI, so feel free to ignore if they do not make sense.

contracts/public-key-registry/src/lib.rs
maksimryndin: yeah, that's the right catch! 👍 
(simple solution would be to have UNIQUE constraint on the table, I think)

contracts/public-key-registry/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+/// Emits one global registration event stream for user key discovery across
+/// all pools in a deployment.
+#[contract]
+pub struct PublicKeyRegistry;
+
+#[contractimpl]
+impl PublicKeyRegistry {
+    /// Register a user's public encryption and note keys.
+    pub fn register(env: Env, account: Account) {
+        account.owner.require_auth();
```
cupicmarko, line 63: Assert length of `encryption_key` and `note_key`.

contracts/public-key-registry/src/test.rs
cupicmarko, line 1: Should we add another test that verifies that specific account can register keys for the address? Or modify this one.

contracts/public-key-registry/src/lib.rs
maksimryndin: good suggestion, Marko 👍

contracts/public-key-registry/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub encryption_key: Bytes,
+    /// BN254 note public key
+    pub note_key: Bytes,
+}
+
+#[contracttype]
+#[derive(Clone, Debug, Eq, PartialEq)]
+enum DataKey {
+    Registration(Address),
+}
```
Fantoni0, line 43: I see we solved the potential duplication on the address book by having a DataKey and storing data on the contract. Not against it, but I think the previous approach was cleaner: smaller contract, no persistent data requirement, cheaper etc. Is there a specific reason this was selected as the solution against handling duplication in the db as originally proposed? What are the tradeoffs?

Other than that, it works, I tried registering multiple times and the issues were solved 🤝

maksimryndin: So, the issue with local resolution of the duplication is less reliable
1) We need at least to enforce a user to wait until the indexer scrapes all the events, and also to check separately that all the events are processed which introduces a specific processing pipeline for this specific event.
2) If we just enforce at the db level, the events have a risk duplicated in the events stream (and then re-played by every app installation)

Fantoni0 (dismissed):
LGTM!

Fantoni0 (approved):
LGTM!

Issue #306, opened 2026-06-26
Check the indexer issue with ahead of RPC


maksimryndin:
see the discussion https://github.com/NethermindEth/stellar-private-payments/pull/298#issuecomment-4810678142

what is the best sync flow? (keeping the background indexing as it is good for the UX)

see also changes
https://github.com/NethermindEth/stellar-private-payments/pull/308/changes
https://github.com/NethermindEth/stellar-private-payments/pull/298/changes/5dcc7384deb7f51e1cb9a5a9f9aad46942d1ac77

Fantoni0:
@v0-e I think this is solved now, right?

v0-e:
That seemed to fix it IIRC :+1:

PR #308, opened 2026-06-26
don't error on ahead indexer

maksimryndin:
- [x] the indexer in the background often moves faster then the latest fetch network tip during an operation. We can omit the error for now as anyway the root check is done further but further investigation is needed https://github.com/NethermindEth/stellar-private-payments/issues/306


v0-e (approved):
lets try it!

PR #311, opened 2026-06-29
SDK init

v0-e:
Adds the main SDK crate (`sdk/pool`, named `stellar-private-payments-sdk`, let me know of other nice and shorter names).

If we could reduce the number of crates to a smaller number it would be nice. Mainly because there is little publish value on some of them and a published crate needs dependencies to be published also.

### Approach

The main struct `PrivatePool` is a self-contained object created per pool. Each `PrivatePool` should only concern with its own notes.

A single-threaded `blocking::PrivatePool` is also provided, using `pollster::block_on` over the main async `PrivatePool` API.

To support both `PrivatePool` impls, and other use cases, we have a few traits:
- `Signer`: e.g., Freighter vs "local" signing;
- `Prover`: e.g., worker abstraction vs direct access;
- `Storage`: same as above. Both to the same SQLite impl.

### Follow ups

- Needs some tests, requires a mock infra;
- sync review: during `execute()` loop if sync is actually required, it awaits for another entity to do that. This is to match and workaround against the app's events sync. Also `PrivatePool::sync()` uses the Indexer which syncs all pools, which may not be desirable;
- `verify_disclosure()` doesn't make much sense to be on `Prover`, though it kind of piggybacks on that due to the loading of circuit artifacts.

Rust SDK 2/3


sdk/pool/src/pool.rs
```diff
@@ -0,0 +1,429 @@
...
+            .await?
+            .pop()
+            .ok_or_else(|| PoolError::Other("transact produced no transaction".into()))
+    }
+
+    pub async fn disclose(
+        &self,
+        req: DisclosureRequest,
+    ) -> Result<Option<DisclosureReceipt>, PoolError> {
+        loop {
```
cupicmarko, line 156: Consider defining a `const DISCLOSE_MAX_RETRIES` or some timeout because it can wait indefinitely.

sdk/pool/src/pool.rs
```diff
@@ -0,0 +1,429 @@
...
+            })
+    }
+
+    pub async fn notes(&self) -> Result<Vec<UserNoteSummary>, PoolError> {
+        self.storage
+            .notes(&self.config.pool_contract_id, &self.config.user_address)
+            .await
+    }
+
+    pub async fn sync(&self) -> Result<(), PoolError> {
```
cupicmarko, line 94: This function syncs every deployed contract, not only the current pool, is that intended?

sdk/pool/src/blocking/runtime.rs
```diff
@@ -0,0 +1,13 @@
...
+
+static RUNTIME: OnceLock<Runtime> = OnceLock::new();
+
+pub(crate) fn block_on<F: Future>(future: F) -> F::Output {
+    RUNTIME
+        .get_or_init(|| {
+            Runtime::new().expect("failed to initialize tokio runtime for blocking SDK API")
+        })
+        .block_on(future)
+}
```
cupicmarko, line 22: Consider something like this because `block_on` will panic if the calling code is already in a Tokio runtime.
```suggestion
pub(crate) fn block_on<F: Future>(future: F) -> F::Output {
    match tokio::runtime::Handle::try_current() {
        Ok(handle) => tokio::task::block_in_place(|| handle.block_on(future)),
        Err(_) => RUNTIME
            .get_or_init(|| Runtime::new().expect("failed to initialize tokio runtime"))
            .block_on(future),
    }
}
```

cupicmarko (commented):
I like the changes, the only thing besides the comments I already made that missing are negative tests for SDK and for PrivatePool blocking.

sdk/pool/src/pool.rs
v0-e: Unfortunately that is the case currently. It uses the Indexer which syncs everything, not specific to the pool. I left the sync review as a follow up, since likely we want also to provide an account level sync (and other methods), and the code here is focused on specific pools.
For an initial integration with the SDK, the web app code will not call `PrivatePool::sync()`, and will rely on the existing sync loop (`events_listener`).

sdk/pool/src/blocking/runtime.rs
v0-e: The current `block_on` follows more or less `reqwest`'s blocking design, which intentionally also panics if used in a Tokio runtime. If the user has an async app, it can just use the non-blocking `crate::PrivatePool` instead.
I'll make this more explicit :+1:

sdk/pool/src/storage/local.rs
```diff
@@ -0,0 +1,130 @@
...
+use crate::{
+    core::process_local_state,
+    disclosure::{DisclosureInputs, DisclosureInputsRequest, map_build_disclosure_inputs},
+    error::PoolError,
+    transact::TransactRequest,
+};
+
+/// In-process SQLite wallet storage (native only).
+pub struct LocalStorage {
+    path: Arc<PathBuf>,
```
maksimryndin: why Arc (threads shareable) for the path and RefCell (only a single thread) for the db?
Also do we have a chance to introduce runtime panics on borrowing when using this with other languages bindings?
(probably, for a followup issue and PR)

sdk/stellar/src/indexer.rs
```diff
@@ -185,22 +212,3 @@ pub trait ContractDataStorage {
         fully_indexed: bool,
     ) -> anyhow::Result<()>;
 }
-impl From<crate::rpc::Event> for ContractEvent {
```
maksimryndin, line 188: I like it :) less code better, more removals! :)

maksimryndin (dismissed):
Outstanding work @v0-e ! 🥇 

Thanks a lot, you've managed that complexity elegantly! 👏 

Some comments but feel free to address in the followups

maksimryndin (approved):

sdk/pool/src/storage/local.rs
v0-e: On `Arc`, it was just a leftover of a previous refactor of having `Storage: Clone`, but that is not needed any more :+1:
On `RefCell`, yes that is an issue :+1: created #322. Mutex is potentially an easy solution.

sdk/pool/src/pool.rs
v0-e: Added that :+1:

sdk/pool/src/blocking/runtime.rs
v0-e: Added an explicit panic message, let me know 🙏

PR #312, opened 2026-06-29
SDK-app first integration

v0-e:
The app now pulls everything through the main SDK crate.

Could wait for the TS SDK, though this already integrates the core SDK code.

### Web app changes

The SDK was integrated. However instead of using the main ops fns like `deposit()`, `transfer()`, the current approach is the core tx plan/execute loop (`prepare_op` -> `loop` over each tx `{ prove, simulate, sign, submit, confirm }`, at `app/crates/platforms/web/src/client/transact.rs`). This is just sprinkle in the `emit_progress` logs to show progress in the UI.

Storage/prover workers were kept.

On the JS interface, the factory style of creating pools was introduced `WebClient.createPool()`. However current changes on the JS app still support only for single pools (e.g., use of `ensureAppPool` fn instead of having a vec/map of pools).

Rust SDK 3/3


app/crates/platforms/web/src/workers/prover.rs
```diff
@@ -161,24 +169,20 @@ async fn load_circuit_artifacts() -> Result<(), JsError> {
...
     let disc_witness_calc =
         WitnessCalculator::new(&disc_wasm_bytes, &disc_r1cs_bytes).map_err(|e| {
             JsError::new(&format!(
                 "failed to init disclosure witness calculator: {e:#}"
             ))
         })?;
-    let disc_prover =
-        Prover::new(DISCLOSURE_PROVING_KEY, &disc_r1cs_bytes).expect("FAILED Disclosure Prover");
+    let disc_prover = Groth16Prover::new(DISCLOSURE_PROVING_KEY, &disc_r1cs_bytes)
+        .expect("FAILED Disclosure Prover");
```
cupicmarko: Is panic here expected?

app/crates/platforms/web/src/client/transact.rs
```diff
@@ -1,319 +1,231 @@
...
-            flow,
-            "sync_check",
-            "Checking sync & ASP membership…",
-        );
+        on_status: Option<Function>,
+    ) -> Result<Option<Vec<String>>, JsError> {
+        let pool = self.inner();
+        let on_status = &on_status;
+        let total = plan.tx_count();
+        let mut hashes = Vec::new();
```
cupicmarko, line 79: `hashes` is returned only if all transactions succeed, if any transaction fails all previous successful transaction hashes are lost.

app/crates/platforms/web/src/client/mod.rs
```diff
@@ -63,102 +82,64 @@ fn emit_progress(
...
-    storage_bridge: OneshotBridge<StorageWorker>,
-    prover_bridge: OneshotBridge<ProverWorker>,
-    fetcher: Rc<CoreStateFetcher>,
+    rpc_url: String,
+    storage: StorageBridge,
+    prover_bridge: ProverBridge,
+    fetcher: Rc<StateFetcher>,
 }
 
 impl Clone for WebClient {
```
cupicmarko: This can be now replaced with #[derive(Clone)]

app/crates/platforms/web/src/workers/prover.rs
v0-e: Mapped to err instead :+1:

app/crates/platforms/web/src/client/transact.rs
v0-e: Now it returns the list of confirmed transactions hashes if the full plan execution fails :+1:

maksimryndin (approved):
LGTM! 😄

PR #323, opened 2026-07-01
Feature/replace witness

Fantoni0:
Replaces ark-circom with a patch of [circom-witness-rs](https://github.com/philsippl/circom-witness-rs). Removing this way the need for wasmer in the app crates. 
Before this, witnesses were generated by JIT compiling the circuit .wasm at runtime via ark-circom + wasmer. This pulled in native Cranelift (conflicting with soroban-sdk, see #192)                                                                                                                      
Now, circuits use a precomputed Circom witness-rs operation graph, committed as a binary blob, to generate the witnesses. 

Thanks to @Kuhai9801 for his exploratory work on #258  :pray: 
This PR is inspired by that work but addresses the issue from a different angle. And makes sure the Circom linkage done by circom-witness-rs stays within the circuit crate to prevent mixing licenses.

It's a draft for now to see if it's worth pursuing it:
- [ ] Behind main, merge and fix conflicts.
- [ ] R1CS has changed, so have the keys (will require a new ceremony)
- [ ] Update docs
- [ ] Remove 2 circom calls  in` build.rs` (see if we can obtain `R1CS` files from circom-witness-rs directly
- [ ] Investigate long `make serve` times. Probably related to the previous point
- [ ] Check if we can remove other ark-circom calls


Fantoni0:
Closing for now.
Completely removing ark-circom is way more difficult than expected. We rely on it multiple times (ceremony, build script, tests), and changing that requires:
1. Write some helper functions (e.g., parsers and readers) that increase the risk of rolling our own stack instead of well-known and tested libraries.
2. Build the generation graphs for circom-witness-rs to work for every circuit we want to test. This significantly increases build times.

Issue #320, opened 2026-06-30
Account-level SDK


v0-e:
Current SDK efforts (e.g., #311) are focused on specific pool logic. However we could benefit from some account level methods, e.g.:
- register public keys;
- balance across pools;
- sync.

Proposed design levels:
- client: `const client = await Client.create({ rpcUrl });`
- account: `const account = await client.account({ userAddress, signer });`. Or merge client/account;
- pool: `const pool = await account.pool({ poolContract });`

PR #331, opened 2026-07-02
Remove leftover SDK `Client`

v0-e:
`Client` was the product of a refactor aimed at introducing a pool client. While we'll likely have a main client in the future (#320), the current is unused and just wrapping `StateFetcher`.
Also updates the top doc example in `lib.rs`.


maksimryndin (approved):
beleza! 🚀

Issue #333, opened 2026-07-02
Consider lazy loading circuits


Fantoni0:
With the support for multiple selective disclosure sizes,  the prover now [loads all artifacts](https://github.com/NethermindEth/stellar-private-payments/blob/8f0d10666b750dd6950a21ef4f34f59aee0d99e8/app/crates/platforms/web/src/workers/prover.rs#L223-L293) when inittialized.
Could we lazy initialize them when the user is in the the disclosure page, or selects more than one note?

maksimryndin:
Probably we should optimize at the level of SDK when an a pool is opening (for both blocking and non-blocking variants) because CLI encounters the same problem - basically for every pool we re-read the same files.

cc @v0-e

v0-e:
I'm hitting frequently the 5 second prover init timeout now. I suppose is related with the new selective disclosure artifacts:
```
16:49:57.905 DEBUG sdk/web/src/workers/prover.rs:289 [WORKER-PROVER] starting... [prover-worker-module.js:865:21](http://localhost:8000/js/stellar-private-payments-sdk-web/dist/workers/prover-worker-module.js)
16:49:57.905 DEBUG sdk/web/src/circuits.rs:46 [circuits] fetching http://localhost:8000/js/stellar-private-payments-sdk-web/dist/circuits/policy_tx_2_2.wasm [prover-worker-module.js:865:21](http://localhost:8000/js/stellar-private-payments-sdk-web/dist/workers/prover-worker-module.js)
16:49:57.905 DEBUG sdk/web/src/circuits.rs:46 [circuits] fetching http://localhost:8000/js/stellar-private-payments-sdk-web/dist/circuits/policy_tx_2_2.r1cs [prover-worker-module.js:865:21](http://localhost:8000/js/stellar-private-payments-sdk-web/dist/workers/prover-worker-module.js)
16:49:57.909 DEBUG sdk/web/src/workers/prover.rs:187 [WORKER-PROVER] fetched policy_tx_2_2.wasm: 646077 bytes [prover-worker-module.js:865:21](http://localhost:8000/js/stellar-private-payments-sdk-web/dist/workers/prover-worker-module.js)
[...]

PR #334, opened 2026-07-02
Cli

maksimryndin:
- [x] cli mimics main flows of the app for UX consistency
- [x] stellar cli handles account/seed management. The only time we expose a secret in memory is with signing but we should handle that later in the issue https://github.com/NethermindEth/stellar-private-payments/issues/71
- [x] legals (licenses, dists, sources are added)
- [x] release GH action
- [x] static binary (to avoid dynamic deps resolution)
- [x] install.sh script to pull the respective binary (and artifacts) from Github releases

to play with it - checkout to the branch `cli` and run

`./deployments/scripts/install.sh --pre`

follow up tasks (to be addressed later, no urgency)
- add disclosure to CLI https://github.com/NethermindEth/stellar-private-payments/issues/339
- add advanced transact flow https://github.com/NethermindEth/stellar-private-payments/issues/338
- add more interim logs at SDK level https://github.com/NethermindEth/stellar-private-payments/issues/337
- SDK blocking should optimize circuits loading otherwise pool opening takes time https://github.com/NethermindEth/stellar-private-payments/issues/333
- https://github.com/NethermindEth/stellar-private-payments/issues/341


Cargo.toml
```diff
@@ -69,7 +75,7 @@ log = { version = "0.4", default-features = false }
 num-bigint = { version = "0.4.6", default-features = false, features = ["serde"] }
 num-integer = { version = "0.1.46", default-features = false }
 reqwest = { version = "0.13", default-features = false, features = ["json"] }
-rusqlite = { version = "0.39.0", default-features = false, features = ["buildtime_bindgen"] }
+rusqlite = { version = "0.39.0", default-features = false, features = ["buildtime_bindgen", "bundled"] }
```
v0-e: We can split the features, i.e., if the web/WASM doesn't need bundled.

maksimryndin: thanks! right call! resolved https://github.com/NethermindEth/stellar-private-payments/pull/334/changes/b710d576f473035d4bcb840ec14c5b2db81e213b

Fantoni0:
Looks good :+1: 
However I experienced some issues and have some comments:
- I think the install script could be in top level `scripts` folder rather than in `deployments/scripts`
- In the same vein, if you install the CLI tool using `cargo install --path ./cli` you'll encounter "File not found" errors when trying to prove something. This is on me, because I missed the install script. But I am sure other people will encounter the same error, maybe we could add a warning "Install using the provided script"
- The main issue: It seems the latest `stellar-cli` tool (v27.0.0) does not expose the secrets `❌ error: Secure Store does not reveal secret key`. So all operations that require it (onboard, register, deposit, transfer, withdraw) will either throw you a warning or directly fail.

v0-e (approved):
Cool, works nicely 🔥 
Stellar-cli based secrets worked fine in my setup (used an account generated in that CLI).

One thing that we have to review however is the sync issue. Multi-tx operations fail mid execution in the CLI / blocking client, because there is no background loop to sync. 
The error is: `Error: ASP membership sync required: SyncRequired(Some(X))`. Basically it is only capable of issuing one transaction at a time.
Though the fix doesn't really pertain in this PR :+1:

Fantoni0:
> Cool, works nicely 🔥 Stellar-cli based secrets worked fine in my setup (used an account generated in that CLI).
> 

Mine was imported from Freighter into `stellar-cli` using `stellar keys add --secure-store MyAccount`
Not sure if relevant, but if it's only me getting the secrets error, it might be important

PR #342, opened 2026-07-03
Rename app `Pool` to `PrivatePool`

v0-e:
A wasm-binden-cli update changed some behavior, causing the current pool app Rust abstraction over the Rust SDK to error out with:
```
2026-07-03T13:41:55.873740Z  INFO downloading wasm-bindgen version="0.2.126"
2026-07-03T13:41:56.017148Z  INFO installing wasm-bindgen
error: class `Pool` referenced by an impl block does not match any exported struct
error: class `Pool` referenced by an impl block does not match any exported struct
error: class `Pool` referenced by an impl block does not match any exported struct
2026-07-03T13:41:56.967607Z ERROR ❌ error
error from build pipeline
```
Site deployment failing because of this.

### DIFF
--- a/app/crates/platforms/web/src/client/disclosure.rs
+++ b/app/crates/platforms/web/src/client/disclosure.rs
@@ -1,12 +1,14 @@
-//! WASM selective-disclosure methods on [`Pool`].
+//! WASM selective-disclosure methods on [`PrivatePool`].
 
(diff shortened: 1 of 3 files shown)


maksimryndin (approved):
a technical term probably is ShieldedPool

but for the marketing purpose, PrivatePool probably is better :)

PR #345, opened 2026-07-03
Increase prover worker init timeout to 20s

v0-e:
Selective disclosure circuits were added, causing the prover initialization to timeout frequently.
More context: https://github.com/NethermindEth/stellar-private-payments/issues/333#issuecomment-4877826244


maksimryndin (approved):
😮‍💨

PR #351, opened 2026-07-06
Auto sync option

v0-e:
Adds option to control when is sync via `SyncMode` enum.
Blocking client uses inline syncs, web client uses background syncs.

### DIFF
--- a/cli/src/cmd/pool.rs
+++ b/cli/src/cmd/pool.rs
@@ -43,15 +43,7 @@ pub fn transfer(
     encryption_key: Option<&str>,
     json: bool,
 ) -> Result<()> {
-    let (session, recipient) = prepare_transfer(
-        config,
-        pool,
-        to,
-        note_key,
-        encryption_key,
-        open,
-        resolve_transfer_recipient,
-    )?;
+    let (session, recipient) = prepare_transfer(config, pool, to, note_key, encryption_key)?;
     let amount = parse_amount(amount)?;
     let results = session
         .pool()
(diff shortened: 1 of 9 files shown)


cli/src/cmd/pool.rs
```diff
@@ -60,26 +52,18 @@ pub fn transfer(
     print_tx_results(config, "Transfer submitted", &results, json)
 }
 
-fn prepare_transfer<C, S, FOpen, FResolve>(
-    config: &C,
```
v0-e, line 64: Also removed these generics.

cli/src/cmd/pool.rs
```diff
@@ -129,43 +113,3 @@ fn print_tx_results(
...
-mod tests {
-    use super::prepare_transfer;
-    use std::{cell::RefCell, rc::Rc};
-    use stellar_private_payments_sdk::{
-        TransferRecipient,
-        types::{EncryptionPublicKey, NotePublicKey},
-    };
-
-    #[test]
-    fn transfer_opens_before_resolving_recipient() {
```
v0-e, line 143: Looks like this test existed to ensure the pool was synced before resolving an address, when that sync happened implicitly on session open; it no longer matters because sync is now explicit on op path.

cli/src/cmd/pool.rs
```diff
@@ -60,26 +52,18 @@ pub fn transfer(
...
-        Option<&str>,
-    ) -> Result<stellar_private_payments_sdk::TransferRecipient>,
-{
-    let session = open_fn(config, pool)?;
-    let recipient = resolve_fn(config, to, note_key, encryption_key)?;
+) -> Result<(PoolSession, stellar_private_payments_sdk::TransferRecipient)> {
+    let session = open(config, pool)?;
+    if matches!((to, note_key, encryption_key), (Some(_), None, None)) {
+        session.sync()?;
+    }
```
v0-e, line 65: This is another awkward sync since we are doing it on pool to resolve a non-pool-specific recipient address.
Also, the base `PrivatePool::transfer` should support address resolution.

maksimryndin (approved):
👍

PR #362, opened 2026-07-07
Defer signing in CLI to Stellar CLI

v0-e:
Uses the Stellar CLI to sign txs instead of importing the secret and creating a local `LocalSigner`.

Also fixes the cause of the error in https://github.com/NethermindEth/stellar-private-payments/pull/334#issuecomment-4892807457 (account added to the Stellar CLI using the `--secure-store` flag). cc @Fantoni0

### DIFF
--- a/cli/src/account.rs
+++ b/cli/src/account.rs
@@ -1,9 +1,9 @@
 //! Account identity resolved from a `stellar keys` alias.
 //!
 //! The account is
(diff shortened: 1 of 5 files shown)


maksimryndin (approved):
Amazing @v0-e 🚀 !

Fantoni0 (approved):
Works great, thanks so much @v0-e 🥳

PR #379, opened 2026-07-09
Configurable pool permissions

### LINKED ISSUE #267: Exclusion list only (open membership) configuration
Allow configurable deployments - e.g. blacklist only.
Having binary switch for the ASP membership list is fine now, no general solution is expected

v0-e:
Provides pools with different permission schemes, 4 variants:
- `open`: without ASP checks;
- `blocklist`: check only ASP non-membership;
- `allowlist`: check only ASP membership;
- `both`: both ASP checks (current deployment). This PR will shift main deployment to `blocklist`, allowing users to fast onboard.

The circuits above wrap the base `policyTransaction.circom`, and `aspMembership.circom` / `aspNonMembership.circom`.

Small change to deployments. use `deploy.sh`, with the pools prefixed with the type above, e.g. for the blocklist-only native token and fully permissioned EURC:
```sh
./deployments/scripts/deploy.sh testnet \
  --deployer <account-id> \
  --asp-levels 10 \
  --pool-levels 10 \
  --max-deposit 1000000000 \
  --pool blocklist:native:$(stellar contract id asset --asset native --network testnet) \
  --pool both:classic:EURC:GB3Q6QDZYTHWT7E5PVS3W7FUT5GVAFC5KSZFFLPU25GO7VTC3NM2ZTVO:$(stellar contract id asset --asset EURC:GB3Q6QDZYTHWT7E5PVS3W7FUT5GVAFC5KSZFFLPU25GO7VTC3NM2ZTVO --network testnet)
```

Also re-deploys. 2nd pool is gated (membership required) since we need to improve token input amounts.

### Related issues
Closes #267.
Adding 3 more circuits, so #333 becomes even more important.


sdk/types/src/policy_tx.rs
```diff
@@ -0,0 +1,55 @@
...
+#[serde(rename_all = "camelCase")]
+pub enum PolicyMode {
+    /// Unrestricted pool transact (`policy_tx_2_2_open`).
+    Open,
+    /// Allowlist only (`policy_tx_2_2_allowlist`).
+    Allowlist,
+    /// Blocklist only (`policy_tx_2_2_blocklist`).
+    Blocklist,
+    /// Allowlist + blocklist (`policy_tx_2_2_both`).
+    Both,
```
maksimryndin: can we make it more future proof?

The mode should be probably an OR bit setting? Or something similar in design.
When unrestricted - None
When allowlist and blacklist - Some(PolicyMode::Allowlist | PolicyMode::Blocklist)

The specific implementation is not important - just it should be additive?

Also on the circuits naming - probably naive idea but can be future proof as well with name enconding
A - allowlist
B - blocklist
S - selective disclosure (or D?)
G - global admin view key.

Then 

policy_tx_2_2_ ... for an open basic one
policy_tx_2_2_A_ ... for an allowlist only
policy_tx_2_2_ABG_ ... for allowlist + blocklist + global admin view key support

As an example.

v0-e: Need some confirmation by @Fantoni0:
- Selective disclosure are a separate set of circuits, right?
- and what about admin view?

I did consider the bit setting initially, though current impl seems more explicit if we don't had more circuit variants to policy_tx.
To confirm, because depending on the case we'll need to provide 2^n (n = flags) circuit variants. And this is not considering different circuits for x notes-in / y notes-out.

maksimryndin: We cannot reliable say we won't have more circuits variants :) that's the point to avoid "Both" (and "Open")

v0-e: Yes, the wording is hard 😄

Fantoni0: Yes, Selective disclosure is a separate circuit. 
Admin view key, will be part of the `policy` circuits, probably as a configurable template, but still part of the main circuit

v0-e: Alright :+1:, so lets go with flags `ABG` like @maksimryndin is proposing? I like it

Deploy example:
```
./deployments/scripts/deploy.sh testnet \
  --deployer <account-id> \
  --asp-levels 10 \
  --pool-levels 10 \
  --max-deposit 1000000000 \
  --pool B:native:$(stellar contract id asset --asset native --network testnet) \
  --pool ABG:classic:EURC:GB3Q6QDZYTHWT7E5PVS3W7FUT5GVAFC5KSZFFLPU25GO7VTC3NM2ZTVO:$(stellar contract id asset --asset EURC:GB3Q6QDZYTHWT7E5PVS3W7FUT5GVAFC5KSZFFLPU25GO7VTC3NM2ZTVO --network testnet)
```

circuits/src/policyTransaction.circom
```diff
@@ -59,32 +30,31 @@ template PolicyTransaction(nIns, nOuts, nMembershipProofs, nNonMembershipProofs,
     signal input outAmount[nOuts];
     signal input outPubkey[nOuts];
     signal input outBlinding[nOuts];
-    
-    // Components and variables definition
+
+    // Exposed to parent wrappers for ASP policy modules; not a main-component output.
+    signal output inPublicKey[nIns];
```
Fantoni0: Since we are making this public (so it can be reused by other circuits), we should add a note here, and probably in the `circuit_keys/README.md` to warn users to NEVER use this circuit as the main one.
- Using it as a component makes `inPublicKey` private (unless stated otherwise).
- Using this directly will reveal the list of public keys.

deployments/scripts/deploy.sh
```diff
@@ -72,18 +98,107 @@ POOL_LEVELS=""
...
+# Sets POOL_SPEC_MODE and POOL_SPEC_BODY for a --pool spec.
+resolve_pool_spec_policy() {
+  local spec="$1"
+  local first="${spec%%:*}"
+  POOL_SPEC_MODE=""
+  POOL_SPEC_BODY=""
+  case "$first" in
+    open|allowlist|blocklist|both)
+      POOL_SPEC_MODE="$(normalize_policy_mode "$first")"
+      POOL_SPEC_BODY="${spec#*:}"
```
Fantoni0: Wouldn't we need to normalize `$first ` before? 
Not sure "OpeN" will pass the `case "$first" in
    open|allowlist|blocklist|both)`

deployments/testnet/circuit_keys/README.md
```diff
@@ -2,16 +2,20 @@
...
-- `testdata/` remains a local/generated workspace directory (and is
-  ignored by git). Tests may still read keys from there.
-- Changing the `policy_tx_2_2` keys requires redeploying the on-chain verifier
-  and any dependent contracts.
-- Changing any `selectiveDisclosure_N` keys requires a web app rebuild and an
-  updated pinned `vk_hash`; it does not require a contract redeploy.
+- `testdata/` remains a local/generated workspace directory (and is ignored by git). Tests may still read keys from there.
+- Changing any `policy_tx_2_2_*` keys requires redeploying the matching on-chain verifier.
+- Changing any `selectiveDisclosure_N` keys requires a web app rebuild and an updated pinned `vk_hash`; it does **not** require a pool contract redeploy.
+- `policy_tx_2_2_both` is the only policy circuit produced by a trusted ceremony (see below).
```
Fantoni0: Not sure about this. The R1CS will have probably changed, so the trusted ceremony keys won't be valid anymore.

Fantoni0 (changes requested):
Looks really good @v0-e . Thanks so much 🙏 , it was a lot of complexity to abstract.
I agree with you on the importance on lazy loading, as we have discussed I faced some timeout issues while testing.

Left some comments. 
Something I noticed, that can probably be a different PR: The UI is not able to render multiple pools for the same asset. So if I want to deploy 3 pools with different policies for the same asset (rather than doing 3 separate deployments), the UI gives you an error

sdk/types/src/policy_tx.rs
maksimryndin: at the CLI level probably having full words is more protective otherwise one letter typo may change the intentions completely :)

What I've meant about the encodings - the file names in the circuits dir.
And for SDK - variants to be additive.

v0-e: Wdyt? 😅 
```
./deployments/scripts/deploy.sh testnet \
  --deployer <account-id> \
  --asp-levels 10 \
  --pool-levels 10 \
  --max-deposit 1000000000 \
  --pool blocklist:native:$(stellar contract id asset --asset native --network testnet) \
  --pool allowlist-blocklist-adminview:classic:EURC:GB3Q6QDZYTHWT7E5PVS3W7FUT5GVAFC5KSZFFLPU25GO7VTC3NM2ZTVO:$(stellar contract id asset --asset EURC:GB3Q6QDZYTHWT7E5PVS3W7FUT5GVAFC5KSZFFLPU25GO7VTC3NM2ZTVO --network testnet)
```

maksimryndin: looks great :)

circuits/src/policyTransaction.circom
v0-e: Updated :+1:

deployments/scripts/deploy.sh
v0-e: This code was replaced with the new flag logic.

deployments/testnet/circuit_keys/README.md
v0-e: Updated :+1:

sdk/types/src/policy_tx.rs
v0-e: Updated to this more composable approach :+1:

Fantoni0 (approved):
Awesome @v0-e . LGTM!

maksimryndin (approved):
tudo otimo! 🥇

PR #380, opened 2026-07-10
Display spent and unspent notes when generating selective disclosures

cupicmarko:
# Display spent and unspent notes when generating selective disclosures

## Description
This PR enables generating selective disclosures for both **spent** and **unspent (available)** notes. 

In private transaction systems based on commitments and nullifiers, spending a note publishes its nullifier to prevent double-spending, but does not remove the note's commitment leaf from the Merkle tree. As a result, the note still exists in the tree and can still be proven to be part of the pool. This change allows users to generate disclosure receipts for spent notes, which is useful for historical audits and proof-of-payment receipts.

Additionally, this PR refactors the frontend to introduce note filtering (by spent status and token/pool) on the disclosure generation screen and standardizes note styling across the application.

### DIFF
--- a/app/js/disclosure.js
+++ b/app/js/disclosure.js
@@ -7,6 +7,7 @@ import {
 } from './wallet.js';
 import { isDbLockedError, showDbLockedModal } from './db-locked.js';
 import { getActivePoolContractId } from './ui/pool.js';
+import { filterNotes, createNoteRow } from './ui/notes-view.js';
 
 // ---------------------------------------------------------------------------
 // Canonical constants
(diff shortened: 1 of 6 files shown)


Fantoni0 (changes requested):
Hey Marko, it works really nice. Just some small comments:
- The All tokens, looks like a button rather than a dropdown list until you click it. Can we show it's a dropdown? 

Image

- I generated a SD proof for 2 unspent and 2 spent notes and the verification said all were unspent

cupicmarko:
>     * The All tokens, looks like a button rather than a dropdown list until you click it. Can we show it's a dropdown?
> 
> 
> Image

Fixed in #393
 
>     * I generated a SD proof for 2 unspent and 2 spent notes and the verification said all were unspent

I didn't reproduce it

Fantoni0 (approved):
LGTM!

PR #381, opened 2026-07-10
Store generated disclosures in persistent storage

cupicmarko:
# Optionally store generated disclosure receipts

This PR introduces an opt-in history feature allowing users to store and manage their generated selective disclosure receipts locally in the application database. Stored receipts are encrypted under the user's own encryption keypair using ECIES, ensuring that sensitive financial disclosures remain confidential while enabling easy auditing, downloading, and verification.

### DIFF
--- a/sdk/prover/src/encryption.rs
+++ b/sdk/prover/src/encryption.rs
@@ -295,14 +295,39 @@ pub fn decrypt_output_note(
 ///
 /// # Returns
 /// Encrypted data (120 bytes)
+/// Encrypt arbitrary-length `plaintext` to `recipient_pubkey` using the same
+/// scheme as note encryption (ephemeral X25519 + XSalsa20Poly1305). Intended
+/// for self-encryption of local artifacts (e.g. disclosure receipts) under the
+/// account's own encryption public key.
+pub fn encrypt_blob(recipient_pubkey: &EncryptionPublicKey, plaintext: &[u8]) -> Result<Vec<u8>> {
+    encrypt_ecies(&recipient_pubkey.0, plaintext)
+}
+
+/// Decrypt bytes produced by [`encrypt_blob`]. Unlike [`decrypt_note_data`]
+/// (which returns an empty vec when a scanned note is not ours), this errors on
+/// failure because the caller expects the blob to be its own.
+pub fn decrypt_blob(
+    recipient_privkey: &EncryptionPrivateKey,
+    ciphertext: &[u8],
+) -> Result<Vec<u8>> {
+    ecies_open(&recipient_privkey.0, ciphertext)
+}
+
 fn encrypt_note_data(recipient_pubkey_bytes: &[u8], plaintext: &[u8]) -> Result<Vec<u8>> {
-    if recipient_pubkey_bytes.len() != 32 {
-        return Err(anyhow!("Recipient public key must be 32 bytes"));
-    }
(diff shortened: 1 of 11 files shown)


sdk/web/src/workers/storage.rs
```diff
@@ -295,6 +310,52 @@ pub(crate) async fn router(req: StorageWorkerRequest) -> Result<StorageWorkerRes
...
+            let id = sha256_hex(receipt_json.as_bytes());
+            let ciphertext =
+                encrypt_blob(&keys.encryption_keypair.public, receipt_json.as_bytes())?;
+            with_storage_mut!(s => s.save_disclosure_receipt(&address, &id, &ciphertext, &created_at)?)?;
+            StorageWorkerResponse::DisclosureReceiptSaved(id)
+        }
+        StorageWorkerRequest::ListDisclosureReceipts(address) => {
+            let keys = with_storage!(s => s.get_user_keys(&address)?)?
+                .ok_or_else(|| anyhow!("user keys not found for {address}"))?;
+            let rows = with_storage!(s => s.list_disclosure_receipts(&address)?)?;
```
Fantoni0, line 329: If one of the rows fails, I think the error will propagate and we won't be able to see any receipts.
So if I have 10 receipts and 1 of them is corrupted, I won't be able to see the other 9?

sdk/state/src/storage.rs
```diff
@@ -16,14 +16,36 @@ use types::{
...
 pub const APP_SETTING_EXPLORER: &str = "explorer";
-
-const MIGRATION_ARRAY: &[M] = &[M::up(include_str!("schema.sql"))];
+pub const APP_SETTING_STORE_RECEIPTS: &str = "store_receipts";
+
+const MIGRATION_ARRAY: &[M] = &[
+    M::up(include_str!("schema.sql")),
+    M::up(
+        "CREATE TABLE disclosure_receipts (
+            id TEXT PRIMARY KEY,
```
Fantoni0, line 25: All other tables define the "ON DELETE CASCADE" in `schema.sql `, is it needed here?

maksimryndin: as we introduce breaking changes anyway (by redeploying contracts - let's directly modify the schema in schema.sql), no need to add a migration atm

sdk/prover/src/encryption.rs
```diff
@@ -295,15 +295,40 @@ pub fn decrypt_output_note(
 ///
 /// # Returns
 /// Encrypted data (120 bytes)
+/// Encrypt arbitrary-length `plaintext` to `recipient_pubkey` using the same
+/// scheme as note encryption (ephemeral X25519 + XSalsa20Poly1305). Intended
+/// for self-encryption of local artifacts (e.g. disclosure receipts) under the
+/// account's own encryption public key.
+pub fn encrypt_blob(recipient_pubkey: &EncryptionPublicKey, plaintext: &[u8]) -> Result<Vec<u8>> {
```
maksimryndin, line 302: for encryption let's come up with a more general solution with https://github.com/NethermindEth/stellar-private-payments/issues/71?

E.g. SQLite has an encryption extensions (maybe we could reuse that?). But as @v0-e introduce a StorageTrait we can think of supporting alternative Postgres, e.g.

maksimryndin (changes requested):

cupicmarko:
Temporarily closed.

PR #385, opened 2026-07-14
New deployment

Fantoni0:
Updates deployment contracts and updpates the db name to be consistent


v0-e (approved):
Nice looking diff on `deployments.json` ✅ 😄

PR #387, opened 2026-07-14
Lazily download circuit artifacts in the WASM SDK

cupicmarko:


v0-e:
Thanks @cupicmarko 🙏 

Unsure if you were planning to do it here, but we'll also need lazy loading of the non-disclosure circuits, we have one now, but will have even more than the disclosure ones.

Also, loading the circuit on operation time is reasonable :+1: but I'm guessing it will add 1-2 seconds of time to the operation. One thing that could be interesting though is loading the circuits when the user enters the respective page/view in the app. A "load circuits" message would need to be added to the prover worker I think. And if that "load circuits" message is never sent, default to loading on operation time like you are proposing here. But is just an idea :+1:

Fantoni0 (approved):
Awesome @cupicmarko .
It works on my Zen 2 CPU 🤟 
Even for multiple-transactions operations, works flawlessly and transparent to the user.

PR #388, opened 2026-07-14
Account SDK

### LINKED ISSUE #320: Account-level SDK

v0-e:
Introduces SDK client and account. Employs a factory design:
- client: network overview/reads, syncs, creates accounts;
- account: address-specific handler, register public keys, creates pools;
- pool: pool ops as before -- some non-pool specific/"broad" methods were moved to client and account.

Basically here the SDK absorbs non-pool specific methods that were implemented independently by the app/CLI. Stuff more app-specific like disclaimer is not included.

Other observations:
- each created object (client -> account -> pool) is independent from its parent. Common resources (storage, prover, signer) are in principle safe to be accessed concurrently (if concurrency is in place);
- while the client and account (and perhaps pool) being split is arguably increasing complexity for the average user, this more open design divides concerns better and allows for more efficient usage by "heavier" users;
- renames `PoolError` to a more crate-wide `Error`. Follow up on increasing its usefulness: #394;
- sync still to be reviewed and follow up. The client shall at least expose a sync loop to run on the background. Optimally for DevEx this could be an inner task, but since we are not making many assumptions on runtime, we can expose a future instead.
- lots of changes and moving around 🙏 and likely some were missed.

### Related issues
Closes #320.


sdk/client/src/pool.rs
```diff
@@ -489,7 +420,7 @@ impl<S: Storage> PrivatePool<S> {
         Ok(results)
     }
```
Fantoni0, line 435: If we execute an operation that takes multiple transactions to happen (requires note merging), and one of them fails. Do we keep state, or know which transactions succeeded and which failed, to only retry some of them?  Or we only propagate the error?

sdk/client/src/account.rs
```diff
@@ -0,0 +1,182 @@
...
+        self.storage
+            .list_user_notes(&self.user_address, limit)
+            .await
+    }
+
+    /// Whether this account's public keys are registered on-chain.
+    ///
+    /// With [`SyncMode::Inline`], local storage is synced before reading.
+    pub async fn is_registered(&self) -> Result<bool, Error> {
+        self.ensure_synced().await?;
```
Fantoni0, line 117: Client, account and pool all have sync() methods, and require ensure_synced(). If I query a specific pool for balance, do all 3 of them force the sync?

sdk/client/src/account.rs
```diff
@@ -0,0 +1,182 @@
...
+            .prepare_register(&self.user_address, note_pk.0, enc_pk.0)
+            .await
+            .map_err(|e| Error::Other(format!("prepare register: {e:#}")))?;
+        let signed = self.signer.sign_soroban_transaction(&prepared).await?;
+        let envelope = TransactionEnvelope::from_xdr_base64(&signed.signed_xdr, Limits::none())
+            .map_err(|e| Error::Other(format!("invalid signed transaction xdr: {e}")))?;
+        let hash = submit_tx(&envelope, fetcher.rpc())
+            .await
+            .map_err(|e| Error::Other(format!("submit register: {e:#}")))?;
+        confirm_tx(hash, fetcher.rpc()).await
```
Fantoni0, line 154: If multiple accounts share the signer and prover, and different pools share the same account. Would multiple calls from the same account to different pools share the same call to `next_sequence` in soroban `tx_prepare`?

cli/src/cmd/feed.rs
```diff
@@ -13,28 +13,11 @@ pub fn run(config: &CliConfig, limit: Option<u32>, json: bool) -> Result<()> {
...
-            Ok(session) => {
-                if let Err(e) = session.pool().sync() {
-                    log::warn!("pool {}: sync: {e:#}", entry.pool_contract_id);
-                }
-            }
-            Err(e) => log::warn!("pool {}: {e:#}", entry.pool_contract_id),
-        }
-    }
+    let session = ClientSession::new(config, &account, &network, true)?;
+    let items = session.operational_feed(limit)?;
```
Fantoni0, line 17: Previously we logged a warn on errors. Would a single fail here get propagated?

Fantoni0 (commented):
Great work @v0-e 
Lots of changes, most for the good. The app seems to be workingflawless, I just left some comments. Most of them are questions for me to understand properly the design.

sdk/client/src/account.rs
v0-e: Ah missed tx sequencing :+1: let me review that. The change should be easy

sdk/client/src/account.rs
v0-e: I'm working on another PR specific for syncing. But yes, currently a sync anywhere syncs everything

Fantoni0: I saw it now, thanks :pray:

cli/src/cmd/feed.rs
v0-e: Is a harder fail now instead of logging just the warn. This a bit of the tradeoff on moving the logic to the SDK, the app/CLI loses a bit "customization". I can bring the warns back, though I'd reason could be slightly better to keep as is as is simpler on the CLI code and to prefer hard fails rather than have partial success. Though let me know 🙏

sdk/client/src/pool.rs
v0-e: On this execute, we only propagate the error :+1:
I think app/WASM side we do provide the ones successful. Let me align this

sdk/client/src/account.rs
v0-e: It was not created yet 😄 should be the branch `ev/client-sync`

Fantoni0: I thought it was the RPC PR :)

cli/src/cmd/feed.rs
Fantoni0: I agree that the harder fail is better. My concern is, if I use the same client and the same account on different pools and one of them fails, the whole call fails? Example: Rather than seeing my balance in all pools (except the failing one), I get a hard fail

v0-e: Gotcha :+1: let me see

sdk/client/src/account.rs
v0-e: Sorry, I thought we were storing the seq num locally. Instead, it is being retrieved from the RPC for each transaction.

To improve concurrency a bit we could store the seq num locally for the account. Every time we issue a tx, we increase it. Though it may not be the best approach if a user is using different apps (e.g., automated trading) to issue txs under the same address. 

Likely we need to allow some customization on tx sequencing.
The problem affects the whole Stellar account.

I can add at least the local seq num or tx sequencing customization (trait) in another PR :+1:. For now I think fetching from RPC every time is currently the safest and more general approach.

cli/src/cmd/feed.rs
v0-e: Functions like `operational_feed()` and `portfolio()` (what you are referring I think, on the balances) now sync everything (all pools) in one call, so we are not calling sync on each pool individually like before. The warns across each pool check then don't really apply.
Wdyt?

sdk/client/src/pool.rs
v0-e: Added partial results to this execute. WASM side we still have a different execute since that has the emit progress calls for the UI.

Fantoni0 (dismissed):
Latest changes look good 🙏

Fantoni0 (dismissed):

Fantoni0 (approved):

PR #395, opened 2026-07-15
fix(bootnode): exempt /healthz and /metrics from RPC rate limiting

adityaNth:
## Description
`GovernorLayer` was applied router-wide, so `/healthz` and `/metrics`shared the same rate-limit budget as the public `/` RPC route. Any burst of real RPC traffic (or even a modest test loop) could exhaust the shared bucket and start 429ing kubelet's probes, causing CrashLoopBackOff.

## Changes
- Scope `GovernorLayer` to only the `/` route instead of the whole router
- `/healthz` and `/metrics` are now never subject to rate limiting, regardless of RPC load

### DIFF
--- a/tools/bootnode/src/http_server.rs
+++ b/tools/bootnode/src/http_server.rs
@@ -60,7 +60,12 @@ impl HttpServer {
         let router = Router::new()
             .route("/healthz", get(healthz))
             .route("/metrics", get(metrics))
-            .route("/", post(handle_rpc))
+            .route(
+                "/",
+                post(handle_rpc).layer(GovernorLayer {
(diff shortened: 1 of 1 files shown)


v0-e (approved):
Thank you @adityaNth! 💯 

I think an approach could also be re-ordering the routes, e.g.:
```rust
Router::new()
    .route("/", post(handle_rpc))
    .layer(GovernorLayer { ... })
    .route("/healthz", get(healthz))
    .route("/metrics", get(metrics))
```
It would cover more routes if we add them in the future.
But of course your solution works beautifully 👍

PR #398, opened 2026-07-16
Re-use RPC client

v0-e:
Re-uses the RPC client instead of creating one almost everywhere.
Cloning it is cheap, allowing an existing connection to be re-used.


Fantoni0 (dismissed):
🤝

Fantoni0 (approved):

PR #400, opened 2026-07-17
SDK background sync

v0-e:
Sync-focused PR:
- moves background sync task to SDK. Users must spawn/run the task themselves using their runtime of choice;
- background sync can now be "kicked", i.e., if we get a sync gap we tell the task to sync instead of having to wait the potential full 5 seconds. Should be more noticeable on multi-tx ops or when the pool has more traffic;
- moves also the bootnode to the sync SDK logic (kept as option). It simply is too much logic to be kept outside of the main sync/SDK, saving the user some trouble.


Fantoni0 (approved):
Looks good! Seems to work in all the tests I did, and now we have the ability of reducing timeouts (as we can directly kick the sync) 👍

PR #403, opened 2026-07-20
Fast indexer catch-up

### LINKED ISSUE #290: Indexer catch-up and limits
Currently the indexer does 10x calls with a 300 event limit each every 5 seconds. Each transact creates 4 events. So if a pool has more than 750 transactions, the indexer will have to at least 5 seconds before we are able to make a transaction.

We can explore a catch-up mode were the indexer does not wait those 5 seconds (or we have a much smaller delay instead), until ledger tip is reached. After we reach sync, we can revert to a slower poll (e.g., 5 seconds).

Additionally, we can increase the 300 event limit per page request, Stellar RPC supports up-to 10000. Increasing the limit would decrease sync time and reduce network strain.

v0-e:
- Avoids waiting 5 seconds while still catching up with ledger tip;
- Bumps fetched events from 300 to 1000 for more efficiency. Stellar RPC allows up to 10000, but likely we cannot assume all node deployments have that max limit.

Closes #290.


Fantoni0 (approved):
Worked flawlessly on a long multi-step withdraw!

PR #405, opened 2026-07-21
Add privacy-safe, configurable tracing across native and web SDKs

### LINKED ISSUE #337: Add more intermediate logs at SDK level
for the blocking SDK purpose I want to see logs for intermediate steps like proving, simulating, submitting etc

cupicmarko:
Replace the ad-hoc log/wasm_log setup with `tracing` end-to-end, close a correlation-ID concurrency bug, redact sensitive fields by tier, wire logging into the CLI, web app, and build tooling, and optimize WASM binary size.

- **sdk/types**: Add a two-tier redaction model — `Sensitive<T>` for amounts/addresses/commitments (redacted unless a debug build explicitly reveals them via `debug_assertions` gate) and `Secret<T>`/key newtypes for keys/seeds/signatures (never revealed, any build). Add the shared `correlation` module: session-prefixed IDs, `CorrelationIdLayer`, and `correlation_id_or_new()` so a function reachable both as a root call (CLI) and as a nested call (web) inherits the caller's ID instead of minting an unrelated one.
- **sdk/client, sdk/web**: Migrate off `log`/`wasm_log` to `tracing`; audit and redact every sensitive log site (client pool ops, storage-worker traces); fix the correlation-ID propagation to ride the span tree instead of a thread-local stack, which broke under concurrent/interleaved async operations.
- **sdk/web/telemetry**: Add `CustomTelemetrySubscriber` directly implementing `tracing::Subscriber` to completely eliminate `tracing-subscriber` from the WASM build graph. Add `TelemetryConfig` (level/sink/ring-buffer/reveal) with privacy-first defaults, a `configureTelemetry` JS API, and custom console formatting that shows span context and correlation IDs in the browser console and ring buffer.
- **Cargo.toml / Cargo.lock**: Strip the `"log"` feature from root `tracing` workspace dependency to remove legacy record shims. Gated `tracing-subscriber` strictly to native targets and dev-dependencies.
- **cli**: Register a tracing subscriber with `CorrelationIdLayer` so SDK spans propagate through CLI-invoked calls; `-v`/`-vv`/`--json` flags remain supported.
- **app**: Wire `configureTelemetry`/`dump_recent_logs` into the frontend facade and add copy/download-diagnostics buttons in Settings.
- **Makefile, Trunk.toml**: Add a `LOGS=1` flag (`make build LOGS=1`, `make serve LOGS=1`) as the primary way to opt into verbose/debug logging; keep `build-debug`/`serve-debug` as aliases. Teach trunk's pre_build hook to build the SDK itself (respecting `LOGS`/`WASM_PROFILE`), tracked via a profile marker so file-watch reloads don't redundantly rebuild.
- **docs**: Document the two-tier privacy model and the logging/build knobs in SECURITY.md, CONTRIBUTING.md, and the crate READMEs.

---

### WASM Binary Size Impact (`make release`)

By implementing a zero-dependency `CustomTelemetrySubscriber` and stripping unused `tracing` compatibility features, the binary footprint overhead of adding end-to-end telemetry and privacy redaction across the SDK is kept to **~59.8 KB (+2.69%)**:

| Artifact | `main` (`c98af40`) | This PR (`b0c6694`) | Size Impact |
| :--- | :--- | :--- | :--- |
| **Bundled Web WASM**<br>`sdk/web/dist/stellar_private_payments_sdk_web_bg.wasm` | **2,219,252 bytes**<br>(2.116 MiB) | **2,279,020 bytes**<br>(2.173 MiB) | **+59,768 bytes**<br>(+58.37 KB / +2.69%) |
| **Raw Un-bindgened WASM**<br>`target/.../release/stellar_private_payments_sdk_web.wasm` | **3,123,045 bytes**<br>(2.978 MiB) | **3,191,425 bytes**<br>(3.043 MiB) | **+68,380 bytes**<br>(+66.78 KB / +2.19%) |

Closes #337 and addresses #368


sdk/client/src/client.rs
```diff
@@ -3,13 +3,34 @@ use types::{ContractConfig, OperationalFeedItem, RecipientLookup};
...
+///
+/// Initialize a native tracing subscriber for CLI or non-WASM consumers.
+///
+/// Reads the `RUST_LOG` environment variable (or defaults to `info`) and
+/// installs a [`tracing_subscriber`] console formatter, plus
+/// [`types::CorrelationIdLayer`] so nested calls can inherit an ambient
+/// `correlation_id` via [`crate::correlation::correlation_id_or_new`].
+/// Subsequent calls are ignored if a subscriber is already installed.
+pub fn init_tracing() {
+    use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
```
v0-e: I'd argue that `tracing-subscriber` should not be on the main SDK crate (library). And therefore also `init_tracing()` could be removed?

sdk/disclosure/src/lib.rs
```diff
@@ -347,6 +347,7 @@ pub fn find_circuit(name: &str) -> Option<&'static RegisteredCircuit> {
 }
 
 /// Validates a receipt against the registered circuit named in the receipt.
+#[tracing::instrument(skip(receipt), fields(stage = "disclosure_receipt_validation", circuit_name = %receipt.circuit.name))]
 ///
 /// # Arguments
 /// * `receipt` - Receipt to validate.
```
v0-e, line 362: We can put the `#[tracing]` attributes after the doc comments to make it more idiomatic.

sdk/tx-planner/src/execute/mod.rs
```diff
@@ -75,6 +75,8 @@ pub struct SpendSession {
 }
 
 impl SpendSession {
+    /// Create a new SpendSession from a wallet, amount, and target.
+    #[tracing::instrument(skip(wallet), fields(stage = "spend_session_setup", wallet_size = wallet.len(), amount = ?amount, target = ?target))]
```
v0-e: `amount` should be `Sensible<>` here also, like in other places?

sdk/tx-planner/src/plan/mod.rs
```diff
@@ -95,6 +95,7 @@ impl<'a> IntoIterator for &'a TransactionPlan {
 }
 
 /// Build a plan from unspent notes and a target spend amount.
+#[tracing::instrument(skip(notes), fields(stage = "transaction_planning", amount = ?amount, note_count = notes.len()))]
```
v0-e: `amount` `Sensible<>`?

sdk/web/src/client/mod.rs
```diff
@@ -100,48 +101,51 @@ impl Client {
...
-            )
-            .map_err(pool_err)?,
-        );
-
-        Ok(Self {
-            storage,
-            inner,
-            prover,
+        with_correlation_id(new_correlation_id(), async {
+            crate::wasm_start();
```
v0-e: Maybe we can add a new `Self::new_inner` with the async work, rather than having a fat closure here? And have the `#[instrument(... fields(correlation_id = %new_correlation_id()))]` attribute there.
Then `new()` calls that.

v0-e (commented):
Very cool 🔥 
We are introducing `tracing` over `log`. The right change, better suited for a product associated with _payments_. A bit heavier though we use few features :+1:. 
Few comments below

cli/src/cmd/feed.rs
```diff
@@ -8,6 +8,12 @@ use crate::{config::CliConfig, explorer::Explorer, onboard, output, session::Cli
 const DEFAULT_LIMIT: u32 = 5;
 
 pub fn run(config: &CliConfig, limit: Option<u32>, json: bool) -> Result<()> {
+    let _span = tracing::info_span!(
+        "cmd_feed",
+        correlation_id = %types::correlation_id_or_new(),
+        limit = ?limit
+    )
+    .entered();
```
v0-e: I think we can prefer having `#[instrument]` attributes if we don't need deep customization of spans.

sdk/client/src/client.rs
```diff
@@ -27,6 +28,8 @@ impl<S: Storage> Client<S> {
         contract_config: ContractConfig,
         bootnode_url: Option<String>,
     ) -> Result<Self, Error> {
+        let _span =
+            tracing::info_span!("client_init", correlation_id = %correlation_id_or_new()).entered();
```
v0-e: Similar comment on preferring `#[instrument]`.

sdk/tx-planner/src/plan/mod.rs
```diff
@@ -105,9 +106,13 @@ pub fn plan(
 
     let values: Vec<NoteAmount> = notes.iter().map(|n| n.amount).collect();
     let combo = find_combination(&values, amount)?;
+    tracing::debug!(?combo, "transaction plan combination result");
```
v0-e: Unsure if really important, but this leaks the excess/change `NoteAmount`s.

v0-e (commented):
Sorry, had these comments on pending state before your last commit 🙏

sdk/client/src/transact.rs
```diff
@@ -274,7 +274,7 @@ fn build_pool_inputs(
         let Some((amount, blinding, leaf_index)) =
             storage.get_unspent_user_note_by_commitment(pool_address, user_address, commitment)?
         else {
-            log::info!(
+            tracing::info!(
                 "unspent note not found for commitment {commitment}; waiting for note derivation"
             );
```
Fantoni0, line 280: Depends on how we define sensitive. But if logs/traces can be shared, I think we would be linking the raw commitment to the user sharing them? Maybe add the Sensitive tag? Or demote from info to debug so it's not exported?

sdk/tx-planner/src/plan/mod.rs
Fantoni0: I think it's better to not leak it

sdk/web/src/workers/storage.rs
```diff
@@ -391,7 +444,7 @@ pub(crate) async fn router(req: StorageWorkerRequest) -> Result<StorageWorkerRes
             StorageWorkerResponse::OperationalFeed(list)
         }
         StorageWorkerRequest::DisclosureInputs(req) => {
-            log::trace!(
+            tracing::trace!(
                 "[{WORKER_NAME}] build selective disclosure inputs for {}",
                 req.user_address
```
Fantoni0: Logging user address. Potentially use sensitive here?

sdk/web/src/telemetry.rs
```diff
@@ -0,0 +1,351 @@
...
+            SPAN_TABLE.with(|table| {
+                table.borrow_mut().insert(
+                    id_val,
+                    SpanData {
+                        correlation_id: visitor.correlation_id,
+                    },
+                );
+            });
+        }
+
```
Fantoni0: Maybe a nitpick, but I am concerned this might grow unbounded for a long-lived tab. Are spans ever closed?

Fantoni0 (commented):
Looks really complete 🚀 
Left some small comments and questions.

cupicmarko:
Rebased to latest main with changes from #412 on this branch
https://github.com/NethermindEth/stellar-private-payments/pull/412#issuecomment-5056688182

sdk/client/src/client.rs
cupicmarko: Good point, I removed it as a dependency

sdk/disclosure/src/lib.rs
cupicmarko: Done

sdk/tx-planner/src/execute/mod.rs
cupicmarko: Done, made `amount` `Sensible<>`

sdk/tx-planner/src/plan/mod.rs
cupicmarko: Done

sdk/web/src/client/mod.rs
cupicmarko: Done, exactly as suggested

cli/src/cmd/feed.rs
cupicmarko: Replaced all manual spans with `#[instrument]` attribute

sdk/client/src/client.rs
cupicmarko: Done

sdk/tx-planner/src/plan/mod.rs
cupicmarko: Fixed.

sdk/client/src/transact.rs
cupicmarko: Fixed with sensitive tag, I think it's better than demotion to debug.

sdk/web/src/telemetry.rs
cupicmarko: It was a real leak, `new_span` insterted per-span data but it was never removed

.cargo/config.toml
```diff
@@ -5,6 +5,9 @@
 [env]
 VERIFIER_VK_JSON = { value = "testdata/policy_tx_2_2_AB_vk.json", relative = true }
 
+[build]
+rustflags = ["--cfg=web_sys_unstable_apis"]
```
v0-e: Do we still need this? Added with gzip changes before I think

v0-e (commented):
Looks good :+1:
Just a couple of more points:
- we show the trace / debug options for logs in the UI, but since we use the release version for the app, they don't come into effect. We can either release with logs or just remove those options from the UI;
- there some "disconnection" of logic from the storage/prover workers on the configuration and copy/download logs. Only the main WASM thread has those features currently.

.cargo/config.toml
cupicmarko: Yes, I reverted the commit, I had by accident committed to local main before I rebased, I missed to revert this single change.

Fantoni0:
My findings are similar to @v0-e , nothing major but:
- The release version degrades logging
```rust
[profile.release]
debug-assertions = false
```

- In `telemetry.rs `, I think that since we inherit the default `tracing_core`, there are some instances when the `enabled()` is not called.

sdk/types/src/logging.rs
```diff
@@ -0,0 +1,257 @@
...
+            assert_eq!(format!("{:?}", val), "\"sensitive_data\"");
+            assert_eq!(format!("{}", val), "sensitive_data");
+        } else {
+            // Production builds must NEVER reveal Tier-1 values
+            assert_eq!(format!("{:?}", val), "\"<redacted>\"");
+            assert_eq!(format!("{}", val), "<redacted>");
+        }
+
+        // Reset state
+        set_reveal_sensitive(false);
```
Fantoni0, line 188: Tests mutate REVEAL_SENSITIVE (global atomic), sometimes this leads to race conditions when running tests in parallel. Telemtry tests use a Mutex to prevent this.

Cargo.toml
```diff
@@ -68,8 +68,8 @@ clap = { version = "4.6.3", default-features = false, features = ["std", "color"
   "error-context",
   "suggestions", "derive", "env"] }
 console_error_panic_hook = { version = "0.1.7", default-features = false }
-futures = { version = "0.3.33", default-features = false, features = ["async-await"] }
-getrandom = { version = "0.2", default-features = false }
+futures = { version = "0.3.32", default-features = false, features = ["async-await"] }
```
Fantoni0: We are regressing from 0.3.33, is this required?

sdk/stellar/Cargo.toml
```diff
@@ -16,8 +16,8 @@ async-trait = { workspace = true }
 base64 = { workspace = true }
 ed25519-dalek = { version = "2.2", default-features = false, features = ["digest", "zeroize"] }
 futures = { workspace = true, features = ["executor"] }
-http = "1.4.2"
-log = { workspace = true }
+http = "1.4.0"
```
Fantoni0: Regressing from 1.4.2. Is this needed?

sdk/state/Cargo.toml
```diff
@@ -16,8 +16,8 @@ types = { workspace = true, features = ["rusqlite"] }
 # external
 anyhow = { workspace = true }
 hex = { workspace = true }
-log = { workspace = true }
-rusqlite_migration = { version = "2.6.0", default-features = false }
+tracing = { workspace = true }
+rusqlite_migration = { version = "2.5.0", default-features = false }
```
Fantoni0: Regressing from 2.6.0. Is this needed?

Fantoni0 (changes requested):
Some minor comments. I think it's almost ready. Thanks @cupicmarko

sdk/state/Cargo.toml
cupicmarko: Restored, all regressed versions come from rebasing, restored to match main.

sdk/stellar/Cargo.toml
cupicmarko: Addressed https://github.com/NethermindEth/stellar-private-payments/pull/405#discussion_r3667424197

Cargo.toml
cupicmarko: Addressed https://github.com/NethermindEth/stellar-private-payments/pull/405#discussion_r3667424197

sdk/types/src/logging.rs
cupicmarko: Fixed, using the same pattern as for telemetry tests.

Fantoni0 (approved):
LGTM!

PR #407, opened 2026-07-22
GVK circuits

### LINKED ISSUE #218: Review BabyJubJub construction and write Circom circuit.
We need to implement the new circuits for global-view key support. We need to enforce in-circuit that the encrypted memos under the admin's public key are indeed the correct encryption of secrets, and not gargage data.

- Review BabyJubJub circuit examples.
- In circuit random derivation and encryption
- Ensure nonce control is secure.
- One for each flavor of global view supported (we can modify the current circuit to provide the vanilla view-only support).
- Manage new circuit keys and make sure artifacts are correctly bundled.

Fantoni0:
Global View Key circuit support.

- Adds the basic circuit for GlobalViewKeyEncryption
- Handles two modes: **view-only** (encrypt outputs only) and **traceable** (encrypt inputs + outputs), selectable per transaction. 
- Composed with all 4 ASP policy configs (Open / Allowlist / Blocklist / Both) * 2 modes = 8 new circuits (A different PR could try to sort better the current circuits organization)
- New docs page: `docs/src/global_view_key.md`

Closes #218 and subissues


Cargo.toml
```diff
@@ -128,5 +127,10 @@ lto = true
 inherits = "release"
 debug-assertions = true
 
+# More aggresive optimization for build scripts
+# Otherwise, compiling the circuit crate takes too much time with new circuits
+[profile.dev.build-override]
+opt-level = 3
+
```
Fantoni0, line 134: As a result of adding 8 new circuits, I had to change the `opt-level`, otherwise the CI always hit the timeout for clippy

v0-e (dismissed):
Amazing 🥇

Fantoni0:
@codex review

circuits/src/globalViewKey.circom
```diff
@@ -0,0 +1,259 @@
...
+    h2.domainSeparation <== 0x05;
+
+    component hr = Poseidon2(3);
+    hr.inputs[0] <== h2.out;
+    hr.inputs[1] <== nonce;
+    hr.inputs[2] <== idx;
+    hr.domainSeparation <== 0x05;
+
+    signal r;
+    r <== hr.out;
```
NiDimi, line 93: Wouldnt it make more sense for the r to come in as a private input? We just need a randomness, and it wont affect anything. I cant think of an attack right now but deriving it deterministically may cause issues in the future. Also more gates for no reason.

circuits/src/globalViewKey.circom
```diff
@@ -0,0 +1,259 @@
...
+    signal output R[2]; // ephemeral public key R = r * G
+    signal output c1;   // encrypted pk
+    signal output c2;   // encrypted amount
+    signal output c3;   // encrypted blinding
+
+    // 1. Validate the authority public key D
+    // D is a non-validated public input. Enforce it satisfies the curve equation.
+    component dCheck = BabyCheck();
+    dCheck.x <== D[0];
+    dCheck.y <== D[1];
```
NiDimi, line 59: I am guessing the contract changes for this will come later, but just a reminder this need to be validated via a public registry. Otherwise nothing is stopping me for supplying my own key. I am sure you are aware but just in case :)

circuits/src/globalViewKey.circom
```diff
@@ -0,0 +1,259 @@
...
+    // BASE8 is the standard Baby JubJub generator of the prime-order subgroup.
+    var BASE8[2] = [
+        5299619240641551281634865583518297030282874472190772894086521144482721001553,
+        16950150798460657717958625567821834550301663161624707787222815936182638968203
+    ];
+    component mulG = EscalarMulFix(254, BASE8);
+    for (var i = 0; i < 254; i++) {
+        mulG.e[i] <== rBits.out[i];
+    }
+    R[0] <== mulG.out[0];
```
NiDimi, line 110: Should we reject R = (0, 1)? If r mod l = 0, then R = S = (0, 1) and anyone can decrypt the memo. Its a nit pick at best because the possibility of this happening is negligible, but given we are aiming for this to be used with high value assets, the safeguard costs us nothing and completely prevents that.

circuits/src/globalViewKey.circom
Fantoni0: Yes, good point. We are aware, but always good to remember. Will add a comment. :ok_hand: 
Still the checks are good just in case the pool-admin deploys a wrong-formed key

circuits/src/globalViewKey.circom
Fantoni0: It is intentional. 
Rather than trusting a user-provided RNG source, we derive it on the circuit. As `r` reuse could leak the encryption's secrets.
It's true that the entire entropy depends on the secrecy of `blinding` (amount, pk can be disclosed/guessed), so I can add an additional secret that acts as salt and further reinforces it.

circuits/src/globalViewKey.circom
Fantoni0: I think it's 1 in (2^254 /8), but it's a good defensive call anyway. Will add some checks to prevent S to be a degenerate point

circuits/src/globalViewKey.circom
NiDimi: I agree that its true, but the blinding is also a user generated random value. So no matter what, either generated in circuit or added as a witness r depends on the RNG source.

Fantoni0: Yes-ish, but it's not the same as it now depends on the nonce, the idx, the blinding and the salt. So we get stronger guarantees on the uniqueness of `r`, and we bind it to the context (without making it depend entirely on it)

v0-e (approved):
:+1:

PR #409, opened 2026-07-22
JS/npm license CI workflow

### LINKED ISSUE #325: Introduce a license linter for js dependencies
see https://github.com/NethermindEth/stellar-private-payments/pull/293#discussion_r3504225469

for Rust we use cargo deny (https://github.com/NethermindEth/stellar-private-payments/blob/main/deny.toml and https://github.com/NethermindEth/stellar-private-payments/blob/main/.github/workflows/dependency-audit.yml#L45), we should have something similar for JS/frontend and a separate github action

cupicmarko:
## Summary

Add a JS/npm license audit pipeline mirroring the existing cargo-deny hygiene for the Rust side, then fix a set of correctness bugs found in review — including one that would have broken the Dependency Review PR gate on every future PR touching a JS lockfile.

## What's here

**Base (`d34e5cf`):**
- `js-license-audit.yml`: a `dependency-review` PR gate (allow-licenses only) plus a `full-tree-scan` job on PRs and a nightly cron, uploading `js-license-scan-report.json`
- `js-license-policy.json`: permissive allowlist mirroring `deny.toml`, a copyleft/UNKNOWN denylist, and target-scoped exceptions for circomlib's GPL-3.0/LGPL-3.0 build tooling and registry-resolved packages
- `check-js-licenses.sh`: full-tree sh+jq scanner enforcing the policy
- `generate-js-attribution.sh`: build-time THIRD-PARTY attribution, wired into `stage-dist-legal.sh`

**Follow-up (`396a782`):**
- **Critical fix**: `dependency-review-action`'s license matcher throws when the allow-list contains a compound SPDX expression (`(MIT AND BSD-3-Clause)`) — the action swallows the exception and treats it as "no match" for *every* dependency in the PR, not just the one with that license. Confirmed by reading the action's source and running its pinned `spdx-license-satisfies@1.0.1` dependency locally. Fixed by stripping compound entries from `allow-licenses` and excusing the one real package that needs it (`sha.js`) via `allow-dependencies-licenses` instead.
- Added a `push` trigger on `main` so the full-tree scan doesn't wait up to 24h (nightly cron) to catch a bad merge.
- `package-circuits-source-bundle.sh` unconditionally `cp -R`'d `circuits/src/`, which would ship circomlib's GPL-3.0 devDependencies (`snarkjs`, `ffjavascript`, etc.) inside the public source bundle if `node_modules` happened to exist on the machine running it. Now pruned defensively — verified by reproducing the leak and confirming the fix closes it.
- `check-js-licenses.sh`: replaced a `echo | while` pipe-subshell with a plain `for` loop (removes shell-dependent ambiguity around `set -e`); a missing target lockfile now fails the scan instead of silently reporting SUCCESS with reduced coverage; an exception whose documented license no longer matches the lockfile's resolved value is now flagged as a violation instead of trusted forever (catches drift from version bumps).
- `js-license-policy.json`: split the `b4a`/`web-worker` exception (b4a's locked version is actually `ISC`, not `Apache-2.0`); corrected the `circomlib` LGPL-3.0 exception's justification, which wrongly claimed the package is build-tool-only when its `.circom` sources are compiled into shipped circuit artifacts; removed an unused `Python-2.0` allowlist entry; fixed allowlist sort order.
- New `scripts/verify-exception-licenses.sh`: a manual maintainer tool that checks every exception's declared license against the npm registry at the package's *locked* version (not `latest`, which can differ after a relicense — this caught the `b4a` bug above). Documented in CONTRIBUTING.md.

## Verification

- shellcheck clean on all shell scripts
- Full-tree scan and attribution generator both still pass end-to-end
- Target-scoped exception isolation, missing-lockfile failure, and stale-exception drift detection all confirmed via crafted fixtures
- The `dependency-review-action` fix confirmed by running the actual pinned library version locally (not just static analysis)

## Notes for reviewers

- All exceptions except the two copyleft ones (the GPL-3.0 circomlib devDependency list and the LGPL-3.0 `circomlib` self-exception) are self-approved after empirical verification against the npm registry; those two remain `PENDING_PR_REVIEW` pending a second look given their real distribution/compliance stakes.
- Consider a legal/compliance review of the LGPL-3.0 source-availability mechanism (`package-circuits-source-bundle.sh` + `stage-dist-legal.sh`) before merge, given it now ships compiled GPL/LGPL-derived circuit logic.

Closes #325


v0-e:
Maybe we can simplify if we ignore `circuits` / `circomlib` analysis?
We do it Rust-side: https://github.com/NethermindEth/stellar-private-payments/blob/main/deny.toml#L18

v0-e (approved):
lgtm!

PR #410, opened 2026-07-22
Shared bootnode DB, fix cursor chains

v0-e:
Bootnode improvements and fixes,
- current deployment was skipping storing empty pagination pages (2-4 KB each) for storage efficiency. At a page per 5 seconds it could easily reach several GBs per month of empty data. However skipping pages breaks cursor chains as a page with events potentially points to (skipped) empty page. Employed solution here is to store the empty pages, however periodically we compress them (sequential empty pages joined into one);
- also, the bootnode now supports different deployments in storage. Helpful in redeployments. Syncing different deployments at the same time still TODO.


tools/bootnode/src/storage/postgres.rs
```diff
@@ -19,27 +26,139 @@ impl Postgres {
...
+        .await?;
+
+    let applied = applied_versions(client).await?;
+    for (version, sql) in MIGRATIONS {
+        if applied.contains(version) {
+            continue;
+        }
+        let tx = client.transaction().await?;
+        tx.batch_execute(sql)
+            .await
```
Fantoni0, line 125: I think if we have 2 fresh instances starting here, both will try to run MIGRATIONS 1, and will require a restart. Can this concurrent fresh start happen?

tools/bootnode/src/rpc.rs
```diff
@@ -81,6 +81,11 @@ impl BootnodeRpc {
         let tip = self.state.ledger_tip.load(Ordering::Relaxed);
         let tip_known = tip > 0;
         let cutoff_ledger = tip.saturating_sub(self.state.cfg.cutoff_ledgers());
+        let in_sync = if tip_known {
+            self.load_in_sync().await?
+        } else {
+            false
+        };
```
Fantoni0: We now call load_in_sync for every getEvents call. Is this needed, or can we try to fetch only on miss?

tools/bootnode/src/storage/postgres.rs
```diff
@@ -19,27 +26,139 @@ impl Postgres {
...
+    }
+
+    async fn list_pages(&self, client: &Client) -> Result<Vec<PageRecord>> {
+        let rows = client
+            .query(
+                r#"
+SELECT id, cursor_in, start_ledger, cursor_out, last_event_ledger, latest_ledger, result
+FROM get_events_pages
+WHERE deployment_id = $1
+ORDER BY id
```
Fantoni0: We select to load all pages, I am afraid it might be heavy after some days running. Is there any way for us to defer the loading?

tools/bootnode/src/indexer.rs
```diff
@@ -65,36 +70,65 @@ impl Indexer {
...
+                    outcome,
+                    InsertPageOutcome::Unchanged {
+                        existing_had_events: false
+                    }
+                )
+            {
+                anyhow::bail!(
+                    "non-empty upstream page not cached for cursor_in={}; \
+                     refusing to advance indexer cursor",
+                    cursor.as_deref().expect("cursor checked")
```
Fantoni0: Is this reachable?

Fantoni0 (commented):
Left some questions. Most of them related to efficiency.
Great work on adding the tests 🙏

tools/bootnode/src/storage/postgres.rs
v0-e: We are just running one bootnode, one DB. The concurrent start can't really happen :+1:

tools/bootnode/src/rpc.rs
v0-e: Improved this to share `in_sync` with the Indexer task. More efficient :+1:

tools/bootnode/src/storage/postgres.rs
v0-e: Indeed :+1: we now load page metadata for the chain walk.

tools/bootnode/src/indexer.rs
v0-e: Nope! It was a leftover :+1:

v0-e:
Thanks for the fast review @Fantoni0 🙌 
Improving compression a bit: now we do compression on the pages that are in (cutoff, tip) range, not < cutoff period since the user may be walking those, and the bootnode should not provide (cutoff, tip) pages.
Added also some more tests

Fantoni0 (approved):
LGTM! Thanks @v0-e

maksimryndin:
amazing @v0-e ! 👍 
let's also remove from the limitations https://github.com/NethermindEth/stellar-private-payments#limitations about no retention or at least adjust the text with the reference to the bootnode(`https://github.com/NethermindEth/stellar-private-payments/tree/main/tools/bootnode`)?

Issue #383, opened 2026-07-10
Verify that we use only SEP interface functions from the wallet


maksimryndin:
https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0043.md
we shouldn't rely on any custom wallet methods

PR #411, opened 2026-07-23
Propagate SEP-0043 user-rejection code (-4) through wallet/SDK error

cupicmarko:
…paths

Wallet rejections previously reached the UI only as free-form message strings, forcing brittle substring matching ("rejected", "denied", ...) at every layer. Thread the SEP-0043 structured error code end to end instead: sdk/web/src/signer.rs and execute/mod.rs classify Freighter JS errors into Error::UserRejected and surface code -4 across the wasm/JS boundary; sdk/web/js/freighter.js and app/js/wallet.js copy that code onto thrown errors; app/js/ui/errors.js and transactions.js key off the structured code first, keeping message matching only as a documented fallback for wallets/versions that don't set it.

Also documents, in app/js/wallet.js and app/ARCHITECTURE.md, why the remaining Freighter-only methods are used instead of their SEP-0043 counterparts — the connect/permission/watch calls pending the Draft standard, and getNetworkDetails as a permanent exception since SEP-0043's getNetwork has no RPC-URL field to replace it (issue #383).


sdk/web/src/signer.rs
v0-e: I think this `rejected` poison all types of signer failures call chains into a user cancel. Maybe we can just change the string?

```suggestion
            .map_err(|e| wallet_js_error(method, "failed", e))?;
```

v0-e (commented):
looks good!

maksimryndin:
awesome @cupicmarko 🚀 
let's also add to the features list https://github.com/NethermindEth/stellar-private-payments#features that SEP-0043 compatible? (yeah, there're caveats you've mentioned but in general?)

Fantoni0 (approved):
LGTM! Thanks @cupicmarko

Issue #340, opened 2026-07-03
[Bug/Test] sdk/witness: missing tests and panic on out-of-range field inputs


Patrick-Ehimen:
### Description

`sdk/witness/src/lib.rs` has two related problems:

### 1. No unit tests

The witness module has zero test coverage despite containing non-trivial logic: JSON parsing, multi-dimensional array flattening, negative-number field encoding, and `BigInt`-to-bytes conversion. The rest of the SDK (`serialization.rs`, `prover`, etc.) has unit tests - the witness module is the exception.

### 2. Panic on out-of-range user input

`to_field_element` (line 114) validates that a value fits within the BN254 scalar field using bare `assert!`:

\`\`\`rust
assert!(abs_value < modulus, "Negative value {} exceeds field modulus", bi);
// ...
assert!(bi < modulus, "Value {} exceeds field modulus", bi);
\`\`\`

This function is called from `flatten_input` and `flatten_pure_array`, both of which process arbitrary JSON supplied by the user via `compute_witness`. A JSON input containing a number ≥ the field modulus - whether from a bug or malformed input  will panic and crash the WASM module instead of surfacing an error to the frontend.

cupicmarko:
Fixed by #413

PR #413, opened 2026-07-23
Fix witness panic on out-of-range inputs, add unit tests (#340)

cupicmarko:
to_field_element validated user-supplied JSON values against the BN254 scalar field modulus with bare assert!s, so an out-of-range number turned into a WASM trap that killed the browser prover worker instead of surfacing an error. It now returns Result and both flatten functions propagate with ?, so malformed input reaches the frontend as a normal error.

Add the crate's first test coverage: 20 unit tests for the pure functions (p-1/p/p+1 boundaries, negative encoding, row-major flattening, indexed and dotted keys, hex/decimal parsing, bool/null coercion, error paths) plus 3 proptest properties for to_field_element (identity in range, p-|v| encoding, Err instead of panic out of range).

The witness_to_bytes asserts are intentionally untouched: they guard internal invariants on computed output, not user input.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -704,7 +704,16 @@ version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
 dependencies = [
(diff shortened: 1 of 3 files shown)


v0-e (approved):
lgtm! 🙌

PR #419, opened 2026-07-24
Remove nullifier map for presence-only check

v0-e:
While testing pools with a larger number of transactions, I was able to reach network read limits (at about ~750 txs).
This PR replaces the nullifier map for a presence-only check, avoiding the load of the entire map on both reads and writes.

### DIFF
--- a/contracts/pool/src/pool.rs
+++ b/contracts/pool/src/pool.rs
@@ -344,21 +344,6 @@ impl PoolContract {
         }
     }
 
-    /// Check if a nullifier has already been spent
-    ///
-    /// # Arguments
-    ///
-    /// * `env` - The Soroban environment
-    /// * `n` - The nullifier to check
-    ///
-    /// # Returns
-    ///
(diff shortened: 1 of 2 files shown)


Fantoni0 (changes requested):
Looks great, in fact, I am not sure why we didn't go with this approach from the beginning.  If I recall correctly, it was because we thought this also implied a Deserealize of the whole key space. It turns out we were wrong. Great job @v0-e 
For the peace of my mind, could we add some tests to ensure the presence/absence of nullifiers still works as expected?

Also, as this is a breaking change and implies re-deploying, what do you think of chaning the visibility of `fn is_spent`?

v0-e:
> For the peace of my mind, could we add some tests to ensure the presence/absence of nullifiers still works as expected?

Added :+1:

> what do you think of chaning the visibility of fn is_spent?

Left it for later, but they are useful for the tests, the correct approach ✅

Fantoni0 (approved):
Awesome!

PR #421, opened 2026-07-27
Pin binaryen version_131 for wasm-opt in sdk build and CI

cupicmarko:
build.sh downloads a sha256-verified binaryen tarball into target/tmp/binaryen/ when the pinned wasm-opt is absent (overridable via WASM_OPT_VERSION, mirroring WASM_BINDGEN_VERSION) and hard-gates on version 131. Selects the release asset by uname -s/-m, covering x86_64/aarch64 Linux and x86_64/arm64 macOS, each with its own pinned sha256; unsupported platforms get a clear error with a WASM_OPT escape hatch to point at a manually installed wasm-opt. Checksum verification falls back to shasum where sha256sum isn't available.

The wasm-build CI job caches the extracted tree keyed on runner.os/runner.arch; the separate install step is dropped since build.sh now self-provisions on a cache miss. Enabler for the wasm-opt size passes; no change to build outputs yet.

Addresses #48

### DIFF
--- a/.github/workflows/wasm-build.yml
+++ b/.github/workflows/wasm-build.yml
@@ -51,6 +51,14 @@ jobs:
       - name: Install SDK npm dependencies
         run: npm ci --prefix sdk/web
 
+      # On a miss build.sh downloads and sha256-verifies binaryen itself; this
+      # only persists the extracted tree between runs.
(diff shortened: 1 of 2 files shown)


v0-e (approved):
lgtm!

PR #422, opened 2026-07-27
Apply pinned wasm-opt -Os to circuit witness WASM in sdk-web build

cupicmarko:
- Run pinned wasm-opt -Os with the six required feature flags on staged circuit witness modules before the sdk-web cargo builds.
- Cache optimized output outside dist/ (target/tmp/witness-opt-cache) keyed by raw input sha256: repeat builds are byte-deterministic and cache files never ship.
- Replace the dist wipe with a selective cleanup preserving circuits/, licenses/, and LICENSE.txt.
- sdk/web/build.rs prefers the staged wasm when it is at least as fresh as the raw artifact and watches BOTH paths unconditionally, so a direct cargo build -p circuits re-triggers the script and the freshness guard is actually evaluated — a stale dist/circuits can no longer win silently.
- Add e2e witness-identity tests asserting byte-identical witness output between unoptimized and optimized modules for all 8 circuits.

Addresses #48


e2e-tests/src/tests/witness_identity.rs
```diff
@@ -0,0 +1,588 @@
...
+    )
+}
+
+#[test]
+fn test_witness_identity_policy_tx_2_2_ab() -> Result<()> {
+    assert_witness_identity(
+        "policy_tx_2_2_AB",
+        &build_policy_inputs(PolicyAspWitness::Both)?,
+    )
+}
```
v0-e: We need to consider then the future variants with GVK. Maybe a single test that covers all policy stems?

v0-e (commented):
Looks good :+1:
Feel like the tests should be under `#[ignored]` or some flag, and not be part of the basic test suite. Especially since they are targeting the web SDK artifacts only and need `make sdk-web-build`.

e2e-tests/src/tests/witness_identity.rs
```diff
@@ -0,0 +1,588 @@
...
+fn workspace_root() -> PathBuf {
+    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .expect("e2e-tests crate must be inside a workspace")
+        .to_path_buf()
+}
+
+fn unoptimized_wasm_path(stem: &str) -> PathBuf {
+    workspace_root()
+        .join("target/circuits-artifacts/release")
```
Fantoni0: Would it make sense to not assume `--release`?
I saw that `sdk/tests/pool.rs` uses this:
```rust
    let profile = std::env::var("PROFILE").unwrap_or_else(|_| "debug".into());
    let circuits = repo.join("target/circuits-artifacts").join(profile);
```

.github/workflows/wasm-build.yml
```diff
@@ -62,6 +62,15 @@ jobs:
       - name: Build npm package
         run: npm run build --prefix sdk/web
 
+      # These tests need both the raw circuit artifacts and the optimized wasm
+      # produced by the sdk-web build above. They are ignored by default and
+      # excluded from Miri runs so other workflows do not try to execute them.
+      - name: Run witness-identity tests
+        env:
+          PROFILE: release
```
Fantoni0: Later in the code, you mention:
```
Cargo does not set `PROFILE` at test runtime, so we use `debug_assertions`
/// as the debug/release proxy.
```
In `sdk/pool/tests`, and resolve to use debug assertions and the helper function `cargo_profile`. So is this needed?

Fantoni0 (commented):
Just a small question

.github/workflows/wasm-build.yml
cupicmarko: Good catch, I didn't check the environment

v0-e (approved):
lgtm!

PR #434, opened 2026-07-28
Replace WASM witness gen with `circom-witness-rs`

### LINKED ISSUE #310: Update Circom optimization level
Right now we build circuits with --O1 [simplification](https://github.com/NethermindEth/stellar-private-payments/blob/main/circuits/build.rs#L293) but we should be using the more aggresive optimization --O2 for production.

It will require creating a new trusted setup/keys for the circuit (as the R1CS will probably change)

v0-e:
This replaces WASM(er) for compiled witness calc graphs computed by Circom C++/`circom-witness-rs`. We further unblocking the publishing of the SDK due to a messy dep tree associated with the `arbitrary` dep. See more at #192.

Performance is also better now. Additionally we switch over to circom `-O2` vs previously `-O1`. Some benches:

|          native    |   now: circom-witness-rs/O2        | prev: wasm(er)/O1 |
| ------------|-------|----|
| avg witness |  0.001s | 0.115s |
| avg groth16 |  0.156s | 0.207s |

|  app/browser   |   now: circom-witness-rs/O2        | prev: wasm(er)/O1 |
| ------------|-------|----|
| avg witness | 0.004s | 0.109s |
| avg groth16 | 1.668s | 2.368s |

With the change to `-O2` we also cut constraints to ~half, reducing also the proving keys size to ~half. This needs a pools redeployment (included here).

Committing the graph bins since compilation time is substantial (measured at ~30min for all circuits). Adding these to git should have low impact since they are just 5%-10% in size in comparison with the committed proving keys. We can explore also committing the R1CS files and provide the full bundle (`git lfs` can be explored).

Run `make witness-graphs` to rebuild the graphs.

### Follow-ups
- graph compilation is a bit ugly: to compile for each circuit one has to call `WITNESS_CPP="<stem>" cargo build -p circuits --features regen-graph` in a loop. While our `circom-witness-rs` fork supports multi-circuit compilations (thanks to @Fantoni0), optimally this change would have to land on one of the "main" witness repos (see next point).
  - Dropped clippy's `--all-features` flag in linter CI, else that would enable the awkward `regen-graph`. This is safe: the few non-default features that we provide are not used. 
  - Related, a nice improvement that could help on this is **moving `circuits/build.rs` logic into its own binary**. Circuit artifacts are such a central part of our process and safety/security, and a dedicated executable could be a more idiomatic design.
- we need an updated `circom-witness-rs` published to publish the SDK. The update to ark `v0.6` is what is critical. Either we can wait for updates in https://github.com/iden3/circom-witnesscalc or https://github.com/philsippl/circom-witness-rs (PRs already open in them) or consider publishing our fork at least temporarily.

------
Adding @Fantoni0, @1sgtpepper as co-authors due to previous efforts. (#323, #258)
Closes #310, #192.


sdk/witness/src/lib.rs
```diff
@@ -1,681 +1,322 @@
...
     #[test]
-    fn witness_to_bytes_roundtrips_full_width_modulus_minus_one() {
-        // p - 1 is the largest valid field element and fills all 32 bytes.
-        let value = modulus() - BigInt::from(1);
-        let bytes = witness_to_bytes(std::slice::from_ref(&value));
-        assert_eq!(bytes.len(), 32);
-        // Reinterpreting the little-endian bytes must recover the original value.
-        let recovered = BigInt::from_bytes_le(Sign::Plus, &bytes);
-        assert_eq!(recovered, value);
+    fn loads_policy_tx_2_2_ab_graph() {
```
cupicmarko: Some `.graph.bin` files are not used in unit tests. Add new tests similar to `loads_policy_tx_2_2_ab_graph` to cover all graphs.

sdk/prover/src/circom_reduction.rs
cupicmarko, line 1: This should be audited line by line, and match to the original implementation (You can check how documentation comments were written in https://github.com/NethermindEth/rs-soroban-ultrahonk for ported logic)

cupicmarko (dismissed):
Great!

sdk/prover/src/circom_reduction.rs
v0-e: Improved the top comment :+1:

sdk/witness/src/lib.rs
v0-e: Testing them all now here :+1:

v0-e: Also we need to add e2e test coverage for this new graph based approach. This is PR keeps the WASM/er tests.
But production/SDK is moving to graph-based proving here.
Added issue #451 to track

app/js/disclosure.js
```diff
@@ -15,13 +15,13 @@ import { App, Toast } from './ui/core.js';
...
+    '0xdd999a5b59411ae38199843aba90fef90fdc4e91ba3169768122efed6c8fcb74',
   selectiveDisclosure_2:
-    '0x5b53adca376d68cd3dc83a02ab9113b3f52cffffe329fdb788d6fe983153584d',
+    '0x937c4ee647d7747f90ec6eff0dcc67f4b00772e8239a87a973267ff15f2dd327',
   selectiveDisclosure_3:
-    '0x46c216ed017af23d5cdd17ce825ebf3180aa3e26481cd2314720f6bac5a49c62',
+    '0x87f63204e0f0106c94fab4aa1a626f7df287604025a3915220624f7166bfaa7c',
   selectiveDisclosure_4:
-    '0xf1346d412fcf9943ccf6774b8648d248918055c68a4d7d9c2a4e417bac5b7cc9',
+    '0x02284c751a2fbdb3f7a731f3dc4a6b3585489d269324c4bcc73afdcbffb66096',
```
Fantoni0: Not sure these are ok. I get `Verification could not be completed: Disclosure receipt circuit metadata mismatch` whenever I verify a SD proof.
Maybe you ran `make witness-graph` again after updating this?

Fantoni0 (changes requested):
Awesome @v0-e , it works way faster now. Just some small questions and comments:

- I see that we are still using ark-circom in some places. And that we still have wasmer in some docs and Cargo.lock. So, the circuit crate still depends on them right, and we still need cranelift?
- Related to the previous one, could we remove the mention to wasmer in deny.toml?

sdk/witness/src/lib.rs
```diff
@@ -1,681 +1,306 @@
...
+    // `bbf_inv(in) = in != 0 ? 1/in : 0` — circomlib `IsZero`.
+    let bbf_inv: BlackBoxFunction =
+        Arc::new(|params: &[BbfFr]| params[0].inverse().unwrap_or_else(|| BbfFr::from(0u64)));
+    bbfs.insert(String::from("bbf_inv"), bbf_inv);
+
+    // `bbf_bit(in, bit) = (in >> bit) & 1` — circomlib `Num2Bits`.
+    let bbf_bit: BlackBoxFunction = Arc::new(|params: &[BbfFr]| {
+        let value = params[0].into_bigint();
+        let bit_index = usize::try_from(params[1].into_bigint().as_ref()[0])
+            .expect("bbf_bit index fits in usize");
```
cupicmarko: This will panic on wasm32 because `usize` is 64bit, `ark-ff` already handles if it's more than 64 bits.

sdk/web/src/workers/prover.rs
```diff
@@ -364,7 +364,8 @@ async fn build_transact_prover(
 
     match fast_path {
         Ok(uncompressed_pk) => {
-            match ProverEngine::new_from_uncompressed_pk(&uncompressed_pk, wasm_bytes, r1cs_bytes) {
```
cupicmarko, line 367: This is a bug I introduced earlier, but it is more noticeable now. `ProverEngine::new_from_uncompressed_pk` accepts `uncompressed_pk` without calling `ensure_sha256_matches`. Disclosure already does this with `fetch_circuit_file_verified`

cupicmarko (dismissed):
LGTM after fixing comments.

cupicmarko:
@Fantoni0 

> Awesome @v0-e , it works way faster now. Just some small questions and comments:
> 
>     * I see that we are still using ark-circom in some places. And that we still have wasmer in some docs and Cargo.lock. So, the circuit crate still depends on them right, and we still need cranelift?

My understanding is that this is only for tooling, but not for witness generation, I updated `CONTRIBUTING.md`. 

>     * Related to the previous one, could we remove the mention to wasmer in deny.toml?

We can't, it will break linter. wasmer is compiled in CI, but it is never shipped.

sdk/web/src/workers/prover.rs
cupicmarko: Fixed, the hashes in `disclosure.js`, `disclosure.md` and README are updated. Also, there is a new new test `vk_hash_matches_committed_keys` that derives tham from both `*_vk.json` and `*_proving_key.bin` files, they can't drift again without test failing.

1sgtpepper:
Some thoughts on the current change: the new witness path still doesn’t preserve Circom’s input contract: unknown and oversized inputs are rejected, but missing or short inputs are accepted and zero-filled. The O1→O2 setting also isn’t part of the artifact freshness check, so persistent OUT_DIR contents can leave an old R1CS in place, while the graph tests still lack a real graph → paired R1CS/PK → VK round trip, currently tracked in #451.

Fantoni0 (approved):
LGTM!
There are some things that can be improved, but it works  and it's a major reduction on proving times. Thanks for raising your feedback @1sgtpepper .
That being said, I think those can be addressed in subsequent PRs, we will probably change how graphs elements are being served. I'll create some issues this week.
To be reviewed by @v0-e next week too.
Thanks everyone who participated 🙏

app/js/disclosure.js
cupicmarko: Fixed.

PR #435, opened 2026-07-29
Pin toolchains and deps, declare MSRV 1.97.1

### LINKED ISSUE #375: Strengthen releases (pin toolchains and deps)

cupicmarko:
## Why

Issue #375. The repo already had a root `rust-toolchain.toml` pin and SHA-pinned actions, but the pinning story was uneven:

- Every workflow passed `toolchain: stable` to dtolnay and relied on the toolchain file *silently overriding* it — works, but confusing, and one missing checkout away from building a release on whatever "stable" means that day.
- Release builds did not pass `--locked`, so the committed `Cargo.lock` was not enforced exactly where reproducibility matters most.
- Several CI-installed tools floated: `trunk`, `mdbook`, `cross` (and `cargo-llvm-cov` in the disabled coverage workflows).
- The bootnode image built on `rust:1.92-bookworm` — a floating minor tag, no digest, and two minors behind the 1.97.1 toolchain pin.
- No `rust-version` (MSRV) was declared anywhere (audit task T10).

## What changed

- **Single source of truth:** removed `toolchain: stable` from all 12 dtolnay steps across 10 workflows; `rust-toolchain.toml` now governs directly. Cleaned up the now-dangling empty `with:` blocks and the stale 1.92.0 comments (`cli-release.yml`, `CONTRIBUTING.md`).
- **Tool pins:** `cross@0.2.5` via taiki-e (both release workflows), `trunk --locked --version 0.21.14` (wasm-build), `mdbook --no-default-features --locked --version 0.5.4` (deployment). These are the current crates.io latest versions — i.e. the pins freeze exactly what the floating installs resolved today, so this is a pure reproducibility gain with zero behavior change.
- **Lockfile enforcement:** `--locked` added to the three release build sites (`cli-release.yml` cross+cargo, `bootnode-release.yml` cross, `tools/bootnode/Dockerfile`). Scoped to release builds only, per issue — not blanket CI.
- **Dockerfile:** builder bumped to `rust:1.97.1-bookworm` and both base images digest-pinned (`rust@sha256:77fac8b9…`, `debian@sha256:7b140f37…`), eliminating the drift from the toolchain pin in the same stroke.
- **MSRV (T10):** `rust-version = "1.97.1"` in `[workspace.package]`, inherited by all 23 members via `rust-version.workspace = true` (verified with `cargo metadata`). New `msrv.yml` workflow.

## Reasoning behind the MSRV design

- **Why 1.97.1 and not `cargo msrv`:** the dependency-declared floor was read from the lockfile via `cargo metadata` — the highest is the vendored `cranelift-control` shim at 1.97.1. Empirically confirmed: `cargo +1.97.0 check --workspace --locked` fails (`cranelift-control requires rustc 1.97.1`), `+1.97.1` passes. No multi-hour `cargo msrv` binary search needed.
- **Why `msrv.yml` is path-triggered** (`**/Cargo.toml`, `**/Cargo.lock`, `rust-toolchain.toml`): the floor can only move when one of those changes, so gating every PR would burn a full second-toolchain compile for zero information (same pattern as `dependency-audit.yml`).
- **Why `override: true` in the job:** the toolchain file would otherwise silently replace the job's toolchain — the project's signature pitfall. The job must check the *declared floor*, not the pin. It's near-vacuous today (MSRV == pin) and becomes load-bearing the day someone bumps the pin: the path trigger fires, and the job proves the floor declaration wasn't forgotten.
- **It only detects, never mutates:** `contents: read`, single `cargo check`, no writes, no PRs — remediation stays a human call.

## Deliberately excluded

- `coverage.yml` / `coverage-pr.yml` — `disabled_manually` on GitHub; the `cargo-llvm-cov` float is a documented known gap, to be handled with the coverage re-enablement issue.
- Nightly rustfmt (`linter`, `bootnode-ci`) and Miri `+nightly` (`ub-detection`) — intentional floats, not pinned by design.
- Upkeep machinery (dispatch-triggered bump workflow + dependabot for `github-actions`, monthly grouped) — follow-up issue.

## Validation

- `cargo check --workspace --locked` green on pinned 1.97.1 and on +1.97.1 explicitly; `cargo metadata`: 23/23 members at `rust_version 1.97.1`.
- All 16 workflows YAML-parse; `toolchain: stable` remains only in the disabled coverage workflows; zero `1.92` references left in the live tree.
- Not run locally: docker build (no daemon) — first exercise is the `bootnode-release` docker job on the next `bootnode-v*` tag.

Closes #375


cupicmarko:
Temporarily closed.

PR #436, opened 2026-07-29
Feature/UI enhancements

cupicmarko:
## What's this PR do?

Adds an in-app confirmation step and a full keyboard flow to deposit, transfer, withdraw, and disclosure-receipt generation — the four actions in the app that either move funds or produce a signed on-chain artifact and can't be undone.

## Why

Freighter already prompts before signing, but that prompt shows the raw transaction/contract-call XDR. For a Soroban pool op carrying note commitments and encrypted keys, that payload is opaque — there's no human-legible amount or recipient in it to check against what you meant to type. So the wallet prompt defends against a forged transaction, not against your own form. It can't catch:

- a mis-parsed amount (`parseAmount` accepted, but not what you intended),
- a withdrawal silently falling back to your own address because the recipient field was left blank,
- a transfer going to the wrong note/encryption key pair.

This PR adds a second, purpose-built confirmation that restates the *parsed* values — not the raw input — right before the wallet is invoked: amount via `Utils.formatTokenAmount`, recipient via `Utils.shortAddress`. It's checking a different thing than Freighter's prompt, not duplicating it.

## What changed

- **`app/js/ui/confirm.js`** (new) — a shared confirmation modal (`confirmAction`), built as plain DOM rather than relying on any framework state, since `ui.js` and `disclosure.js` are separate esbuild bundles with no shared module graph. Cancel is focused by default (safe default: a stray Enter cancels, not confirms); Confirm is reachable by Tab or click. Escape/overlay-click/Cancel all resolve `false`; the promise settles exactly once.
- **`app/js/ui/keys.js`** (new) — a tiny shared `onEnter` helper, split out for the same cross-bundle reason as `confirm.js`, replacing two copy-pasted implementations.
- **`app/js/ui/transactions.js`** — deposit/transfer/withdraw now route through `confirmAction` before submitting, from both the button click and Enter in the amount field. Enter also advances between fields (address → amount) instead of only submitting on click.
  - Enter bypasses the submit button's `disabled` state, so a naive Enter-to-submit binding could double-fire mid-transaction; `onEnterUnlessBusy` guards against that explicitly.
  - Recipient/transfer-address fields are now validated with `StrKey.isValidEd25519PublicKey` before proceeding, so a malformed address fails fast with a toast instead of reaching the confirmation screen.
  - Fixed the transfer-address Enter handler to skip the registry lookup when the `input` handler already resolved it, and to focus the note-key field (not the amount field) when no registration was found and manual entry is required.
- **`app/js/disclosure.js`** — same pattern for the receipt-generation form: Enter advances through authority → payload → purpose → nonce, and Enter in the nonce field runs the same confirm-then-generate path as the button. A local `generating` flag guards against double-submit, since the generate button's `disabled` state is also overloaded to mean "invalid note selection" and can't be reused as the sole in-flight guard.
- **`app/js/ui/navigation.js`** — the settings panel now closes on overlay click or Escape, matching the dismiss behavior just added to the confirm modal so the two overlays behave consistently.

## Risk / tradeoffs

- This adds one extra click (or Enter) before every deposit/transfer/withdraw/disclosure-generate, on top of the existing wallet signing prompt. That's a deliberate tradeoff: irreversible actions get a human-readable checkpoint even though it costs a step.
- The keyboard flow and the confirm gate are both new interaction surfaces; recommend a manual pass through all four flows (deposit, transfer with and without an existing registry entry, withdraw with a blank vs. explicit recipient, disclosure generate) before merging.


Fantoni0 (dismissed):
I have mixed feelings here to be honest.
I understand the motivation of the PR, and I agree that it's better to provide a confirmation with readable info to the average user. But I am not a big fan of the UX change.
My proposal: we merge as it is, and @cupicmarko you create some issues for contributors to help with:
- Confirmation does not appear in advanced mode. Not sure if intentional, as you assume the advanced user knows what he is doing.
- Confirmation could show some additional information about the transaction (e.g. fee, number of required transactions for a multi-step action)
- Not sure if we need confirmation in Selective Disclosure, as it's purely an off-chain action and requires the user to send the proof by himself.
- Any other idea you might have on extending the confirmation.

In general, I think it's good as it's additional protection for the user.

cupicmarko:
> I have mixed feelings here to be honest. I understand the motivation of the PR, and I agree that it's better to provide a confirmation with readable info to the average user. But I am not a big fan of the UX change. My proposal: we merge as it is, and @cupicmarko you create some issues for contributors to help with:
> 
>     * Confirmation does not appear in advanced mode. Not sure if intentional, as you assume the advanced user knows what he is doing.


I didn't commit anything in the advanced mode because I didn't test it properly.

>     * Confirmation could show some additional information about the transaction (e.g. fee, number of required transactions for a multi-step action)

OK, this is a really nice addition, good thinking 

>     * Not sure if we need confirmation in Selective Disclosure, as it's purely an off-chain action and requires the user to send the proof by himself.

Yeah, that was a cheap edit and completely useless.

Fantoni0 (approved):
LGTM! Thanks @cupicmarko

Issue #279, opened 2026-06-18
Define Traceable view key Memo


Fantoni0:
For the traceable disclosure memo, both inputs and outputs are encrypted:
- [ ] Extend structures to also carry encrypted note inputs under the global VK.

Issue #278, opened 2026-06-18
Define Vanilla view Key memo


Fantoni0:
For vanilla disclosure, each new output additionally includes an encrypted version of its secrets for the global view-key authority (Simmilarly as it includes an encrypted memo for note discovery)

- [ ] Extend output/input management structures to carry the vanilla ViewKeyMemo.
- [ ] Define how the memo is attached to each output (1-to-1 relation)

Issue #277, opened 2026-06-18
Define ViewKeyMemo


Fantoni0:
Define `ViewKeyMemo` type in `app/crates/core`. Probably a new "viewkey" member in the workspace. Initially, just a pure data struct + serialization.
Similarly to what we did for new Disclosure types.

PR #437, opened 2026-07-29
Gvk new memo

### LINKED ISSUE #219: New encrypted memo type
As a result of the global view key, we'll need to update the input/output management structure. Both in the Rust crates and on the smart contract. This issue focuses on the off-chain components.

- Define new ViewKeyMemo.
- For vanilla disclosure, now each new output includes additionally an encrypted version of its secrets for the global view key authority.
- For traceable disclosure, we now also encrypt note inputs under the global view key.

Fantoni0:
Small PR to support the new encrypted memos required by global view key support:
- Creates new `sdk/types/gvk.rs`.
- Supports view-only and traceable memos.
- It's backward compatible with all previous policy configurations.

Closes #219 #277 #278 #279

### DIFF
--- a/sdk/types/src/lib.rs
+++ b/sdk/types/src/lib.rs
@@ -50,6 +50,13 @@ pub struct PoolConfigEntry {
     pub asset: AssetDescriptor,
     /// ASP policy flags for transact proofs.
     pub policy_flags: PolicyFlags,
+    /// Global View Key mode for this pool. Defaults to [`GvkMode::Off`] for backwards compatibility
+    #[serde(default)]
+    pub gvk_mode: GvkMode,
+    /// Pool administrator's Baby JubJub public key `D`, informational only
+    /// not verified. `None` when `gvk_mode` is [`GvkMode::Off`].
+    #[serde(default)]
+    pub gvk_authority_pub_key: Option<BabyJubJubPoint>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/sdk/types/src/lib.rs
+++ b/sdk/types/src/lib.rs
@@ -2,12 +2,14 @@ mod amounts;
 mod chain_data;
 mod disclosure;
 mod ext_data;
+mod gvk;
 mod policy_tx;
 pub use amounts::*;
 use anyhow::{Result, anyhow};
 pub use chain_data::*;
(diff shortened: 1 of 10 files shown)


sdk/types/src/lib.rs
```diff
@@ -581,3 +591,94 @@ impl_byte_wrapper!(EncryptionPrivateKey);
...
+                    "kind": "classic",
+                    "code": "EURC",
+                    "issuer": "GB3Q6QDZYTHWT7E5PVS3W7FUT5GVAFC5KSZFFLPU25GO7VTC3NM2ZTVO"
+                }
+            }
+        ]
+    }"#;
+
+    #[test]
+    fn legacy_pre_gvk_deployments_json_parses_with_gvk_off() -> Result<()> {
```
v0-e: Wonder if we need legacy config tests? JSON is additive, and we are still in heavy development so we can break stuff anyway? 😄 
Just a nit

sdk/types/src/lib.rs
```diff
@@ -52,6 +54,14 @@ pub struct PoolConfigEntry {
     pub asset: AssetDescriptor,
     /// ASP policy flags for transact proofs.
     pub policy_flags: PolicyFlags,
+    /// Global View Key mode for this pool. Defaults to [`GvkMode::Off`] for
+    /// backwards compatibility
+    #[serde(default)]
+    pub gvk_mode: GvkMode,
+    /// Pool administrator's Baby JubJub public key `D`, informational only
+    /// not verified. `None` when `gvk_mode` is [`GvkMode::Off`].
+    #[serde(default)]
```
v0-e: We can put here `skip_serializing_if` when `Option::None`, else it will be serialized as `null`.
Just a nit, the opposite could be preferred also ofc

v0-e (dismissed):
lgtm! just some skippable nits

sdk/types/src/lib.rs
Fantoni0: You got me at 
> so we can break stuff anyway

Fair point

sdk/types/src/lib.rs
Fantoni0: Good call

v0-e (approved):
lgtm!

PR #440, opened 2026-07-30
fix(ci): bump actions/cache to v6.1.0

cupicmarko:
- GitHub rejected the pinned v4.0.2 SHA (0c45773b) in.yml as a deprecated actions/cache version. Update both deployment.yml and wasm-build.yml to the v6.1.0 SHA (5a3ec84e). Only the stable path/key inputs are used, which are unchanged across v4 → v6, and the runner (2.336.0) meets the v5+ minimum requirement.

- Run cargo sort


v0-e (approved):
lgtm!

Fantoni0 (approved):

PR #446, opened 2026-07-30
Opt-in Freighter to `stellar-private-payments/freighter`

v0-e:
The main package entry always re-exported `FreighterSigner`, so importing anything from `stellar-private-payments` also loaded `@stellar/freighter-api`.
- That only ships UMD, which breaks under native ESM.
- Now we keep Freighter on the opt-in `stellar-private-payments/freighter` subpath so the core SDK no longer pulls it in by default.


Fantoni0 (approved):
Awesome catch! @v0-e

PR #447, opened 2026-07-30
feat: Default bootnode URL in CLI onboarding

### LINKED ISSUE #430: CLI: Add default bootnode configuration
The CLI tool's onboarding process does not provide the default bootnode URL `https://bootnode.dev-nethermind.xyz` (as the webapp does). 

Run `spp onboard --account YourAccount` to check that by default it simply leaves it as an empty string. It should provide the default URL

PoulavBhowmick03:
## 🚀 What’s this PR do?

- `configure_bootnode` now shows the default in the prompt and treats a blank
  answer as "accept", mirroring `configure_explorer`.
- `get_bootnode_setting` defaults to `enabled: true` with the default URL when
  nothing is stored, mirroring `explorer::base_url`.

---
Closes #430 
---

## 🧠 Context

`spp onboard` previously left the bootnode URL empty by default, while the webapp
filled it with `https://bootnode.dev-nethermind.xyz`.

---
## ✅ Required Checklist

- [X] I’ve read the [contributing guide](../CONTRIBUTING.md).
- [X] I’ve tested this locally, run .githooks/pre-push and provided test instructions for maintainers if needed
- [X] I’ve added relevant docs or comments
- [X] I’ve updated or created tests if needed


cli/src/onboard.rs
```diff
@@ -165,17 +165,16 @@ fn configure_bootnode(
     }
     let current = storage.get_bootnode_setting()?;
     println!("\nBootnode fallback (optional):\n{BOOTNODE_RISKS}\n{BOOTNODE_TEXT}");
-    let hint = if current.enabled && !current.url.is_empty() {
-        format!(" [{}]", current.url)
+    let hint = if current.url.is_empty() {
+        DEFAULT_BOOTNODE_URL
     } else {
-        String::new()
+        &current.url
```
Fantoni0: We are forcing the user to trust the default bootnode. With no warning. 
The user must be able to skip/disable bootnode if he does not want to accept the bootnode risks.

sdk/state/src/storage.rs
```diff
@@ -366,8 +367,8 @@ impl Storage {
         Ok(self
             .get_setting_json(APP_SETTING_BOOTNODE_CONFIG)?
             .unwrap_or(BootnodeSetting {
-                enabled: false,
-                url: String::new(),
+                enabled: true,
+                url: DEFAULT_BOOTNODE_URL.to_string(),
```
Fantoni0: Same as above

Fantoni0 (changes requested):

cli/src/onboard.rs
PoulavBhowmick03: i thought about this yesterday but went with it enabled for all the times. should it be like the disclaimer? where the user prompts with a yes/no on if they want the default bootnode_url, with the `BOOTNODE_RISKS`, if they say yes, and next step if they keep it blank, the default is selected, and if they say no, its disabled. or is there another way of doing it?

Fantoni0: Rather than what we have now: 
image

We could have `Bootnode archive URL (enter to accept): https://bootnode.dev-nethermind.xyz ` as a prefilled URL text the user can edit or accept.

If he deletes and provides no URL, we don't configure the bootnode (as it happened before). But is the user wants to use the default, it's just an enter for him.

PoulavBhowmick03:
Hi @Fantoni0, given the interactive prefill prompt you mentioned, i have added that using dialoguer for the initial url render and edit/removal of it. ptal

cli/src/onboard.rs
```diff
@@ -166,15 +166,15 @@ fn configure_bootnode(
...
     println!("\nBootnode fallback (optional):\n{BOOTNODE_RISKS}\n{BOOTNODE_TEXT}");
     let hint = if current.enabled && !current.url.is_empty() {
-        format!(" [{}]", current.url)
+        &current.url
     } else {
-        String::new()
+        DEFAULT_BOOTNODE_URL
     };
-    let input = prompt_line(&format!("Bootnode archive URL{hint} (blank to skip): "))?;
+    let input = prompt_line_prefill(&format!("Bootnode archive URL (blank to skip)"), &hint)?;
```
Fantoni0: I see why you did this. And if there were more instances, it would be the right thing to do. But i think for now it's better to:
1. Remove the dialoguer dependency
2. Use the same approach as the function `configure_explorer`. It's simpler, and it will work

cli/src/onboard.rs
```diff
@@ -166,15 +166,15 @@ fn configure_bootnode(
...
     if !input.is_empty() {
         storage.set_bootnode_setting(true, &input)?;
-    } else if !current.enabled {
-        storage.set_bootnode_setting(false, "")?;
+    } else {
+        storage.set_bootnode_setting(true, &input)?;
```
Fantoni0:
```suggestion
        storage.set_bootnode_setting(false, "")?;
```

Fantoni0 (changes requested):
Thanks so much @PoulavBhowmick03  💯 
It's close, I just left some small comments. Let's see if we can do it without the additional dependency.

cli/src/onboard.rs
PoulavBhowmick03: got it. making the change and not using dialoguer

PoulavBhowmick03:
@Fantoni0 added a "none" option in the prompt so that if user enters that, it disables the bootnode_url and empty sets to default.

Fantoni0 (approved):
LGTM!
Thanks @PoulavBhowmick03 🥳

PR #449, opened 2026-07-31
ci(license): remove dependency-review, ban fastrlp

cupicmarko:
The dependency-review action in the JS license audit workflow
 duplicated coverage that cargo-deny already provides for Rust/Cargo
 dependencies via dependency-audit.yml. Remove the redundant job and
 add a comment clarifying the split.

 Also deny the MPL-2.0 fastrlp crate in deny.toml; it is only pulled
 in by an unused feature gate and must never be part of the compiled
 dependency graph.


v0-e (approved):
cool!

PR #450, opened 2026-07-31
Merge SDK crates

v0-e:
New SDK crate structure for SDK publishing (no local dependencies):
- `sdk/native`: base Rust SDK crate, with most of previous `sdk/` crates under it as modules;
- `sdk/web`: same as before;
- `sdk/tests`: same as before.

Also takes out the `-sdk` suffix on most places. E.g., the Rust SDK crate will be released as `stellar-private-payments`.


sdk/tx-planner/README.md
```diff
@@ -1,49 +0,0 @@
-# tx-planner
-
-Builds a **transaction plan** for private pool spends: which wallet notes to use, how many on-chain `transact` calls to make, and what each step does.
-
-Each on-chain transaction is a **2-in / 2-out** `transact` (at most two real inputs, two outputs). The planner only emits steps that fit that shape.
-
-## What it optimizes for
-
-The goal is to pay a target `NoteAmount` while keeping wallet and chain cost low:
-
```
Fantoni0, line 10: I think this is lost in the refactor. We don't have a README for tx-planner anymore

sdk/web/build.rs
```diff
@@ -3,7 +3,7 @@ use std::{
     env, fs,
     path::{Path, PathBuf},
 };
-use types::PolicyFlags;
+use stellar_private_payments::types::PolicyFlags;
```
Fantoni0, line 6: Not sure if it could be a problem, but are w compiling the whole SDK (compared to only the types crate) now?

tools/bootnode/Cargo.toml
```diff
@@ -16,7 +16,7 @@ path = "bin/bootnode.rs"
 
 [dependencies]
 # workspace
-types = { path = "../../sdk/types", version = "0.1.0" }
+stellar-private-payments = { path = "../../sdk/native", version = "0.1.0" }
```
Fantoni0, line 19: Similar to the previous comment. Now we pull all the crates, not only types?

sdk/native/Cargo.toml
```diff
@@ -0,0 +1,96 @@
...
+description = "Stellar Private Payments Rust SDK"
+build = "build.rs"
+
+[lib]
+name = "stellar_private_payments"
+
+[features]
+default = ["rusqlite"]
+# Kept for source compatibility with former types crate cfgs; always enabled in this fat crate.
+rusqlite = []
```
Fantoni0: How does this work?

sdk/native/src/lib.rs
```diff
@@ -42,47 +42,12 @@
...
-        APP_SETTING_BOOTNODE_CONFIG, APP_SETTING_EXPLORER, AccountKeys,
-        CURRENT_DISCLAIMER_HASH_HEX, CURRENT_DISCLAIMER_TEXT_MD, DEFAULT_BOOTNODE_URL,
-        DerivedUserNoteRow, PoolCommitmentRow, SqliteStorage, StoredUserKeys, process_events,
-        process_notes,
-    };
-}
+pub mod planner;
+pub mod state;
+pub mod types;
+pub mod zk;
```
Fantoni0, line 50: We went from specific re-exports to exposing all the surface

Fantoni0 (commented):
Thanks @v0-e !
Looks mostly good, all test pass. Just some question/comments I left.

- There are still some old mentions to deleted paths. Maybe running a quick grep.
- The crate got bigger, and the exposed surface too. If it's required and it's a tradeoff it's fine, but wanted to mention it.

sdk/tx-planner/README.md
v0-e: Indeed, resurrected it :+1:

sdk/web/build.rs
v0-e: Yeah, but no big deal for now I'd say. However, one thing that will be interesting is adding some feature flags to the SDK to reduce pulled deps / compile times for some use-cases :+1: (#492)

tools/bootnode/Cargo.toml
v0-e: We pull one crate now, it is bigger though 😄 
Feature flags can be an interesting approach to reduce this :+1:

sdk/native/Cargo.toml
v0-e: An hallucination, but can be helpful in the future. Removed it :+1:

sdk/native/src/lib.rs
v0-e: Yeah, that deserves a review, but not so easy as we may limit some use-cases unknowingly. (#493)

Fantoni0 (approved):
We can merge as it is, and address the surface reduction and the introduction of features in later PRs

PR #453, opened 2026-07-31
feat(sdk): add Rust SDK usage examples

cupicmarko:
Add six blocking-client examples (account_pool, sync, estimate, deposit, transfer, withdraw) plus a shared common/bootstrap module for env-var handling, client construction, and artifact loading.

Also document the examples in sdk/client/README.md and add a CI step to compile them.


sdk/client/examples/SETUP.md
```diff
@@ -0,0 +1,447 @@
...
+Use the local bootnode during onboarding (`--bootnode-url
+http://127.0.0.1:8080`) and when running the examples by setting
+`SPP_BOOTNODE_URL=http://127.0.0.1:8080`.
+
+## ASP membership for allowlist pools
+
+The default testnet pool is configured with an `allowlist` policy flag. Before a
+wallet can `deposit` or `transfer` through an allowlist pool, the pool admin
+must insert that wallet's ASP membership leaf into the `asp_membership`
+contract.
```
Fantoni0: Not sure this is true, the default XLM is open for anyone to interact with (no allow or blocklist). It would be great to allow the examples without having the ASP secret configured for this scenario. Also, setting the ASP secret can be a bit of work, so it'd be nice to only require it if needed.

sdk/client/examples/transfer.rs
```diff
@@ -0,0 +1,167 @@
...
+    println!("Recipient: {recipient:?}");
+
+    println!();
+    println!("Estimating transaction count...");
+    let plan = pool.prepare_transfer(&notes, recipient.clone(), amount)?;
+    println!("Expected on-chain transactions: {}", plan.tx_count());
+
+    println!();
+    println!("Submitting transfer (proving may take a while)...");
+    match pool.transfer(recipient, amount) {
```
Fantoni0, line 105: I think transfer internally calls prepare_transfer? Or is this intentional to print the expected transactions?

sdk/client/examples/deposit.rs
```diff
@@ -0,0 +1,159 @@
...
+fn print_result(result: &stellar_private_payments_sdk::types::TransactionResult, pool_id: &str) {
+    println!("Deposit submitted and confirmed.");
+    println!("  tx hash:   {}", result.tx_hash);
+    println!(
+        "  explorer:  https://stellar.expert/explorer/testnet/tx/{}",
+        result.tx_hash
+    );
+    println!();
+    println!(
+        "Check the updated pool balance with: cargo run --example account_pool -- SPP_POOL_CONTRACT_ID={pool_id}"
```
Fantoni0: Is this equivalent to `SPP_POOL_CONTRACT_ID={pool_id} cargo run --example account_pool` ?

Fantoni0 (changes requested):
Looks really good @cupicmarko 
Left some questions

sdk/client/examples/SETUP.md
cupicmarko: You're right. It was the opposite. Fixed

sdk/client/examples/transfer.rs
cupicmarko: Yes, it was intentional, the call is there only to print the expected transaction count before commiting to the real submit.

sdk/client/examples/deposit.rs
cupicmarko: It was intended to be equivalent, but the printed command was wrong

Fantoni0 (dismissed):
Once conflicts are solved, LGTM!
Thanks @cupicmarko , really comprehensive and great documentation in the SETUP.md. It's not trivial, but the guideline it's quite helpful

Fantoni0 (approved):

PR #454, opened 2026-07-31
Bootnode sealed empty tip healing

v0-e:
Fix a client-stuck scenario: a terminal empty page (tip self-loop, indexer in sync) whose `latest_ledger` is below the retention cutoff used to hand off immediately. Wallets that advance sync position `from latest_ledger` on empty pages could not progress. The fix introduces a two-step flow: heal once, then hand off.


cupicmarko (approved):
Looks good!
Can we add an integration test for new heal path, probably in new PR?

v0-e:
Yes integration tests with the client are the best idea 💯

Issue #280, opened 2026-06-18
Research event versioning


Fantoni0:
We'll need to manage multiple versions of `newComitmment` on different pools:

- Pool with no view key
- Pool with vanilla view key
- Pool with traceable view key

- [ ] Research how to manage this on the indexer side. Take into account that we might need to add new types in the future.

Issue #233, opened 2026-05-28
Abstract Verifiers


Fantoni0:
In a similar way to how #206 handled multiple pools we need to:

- Handle multiple artifacts/keys for circuits.
- Manage multiple verifiers (at most one per pool)  for different pools that implement different features. With the modularity we have, each pool can use a different circuit. This means different keys and different verifier contracts that need to be managed.
- Manage automatic switch of pools in the UI to trigger them.

Related to #223

maksimryndin:
this task should wait for some time until we accumulate enough cases

PR #461, opened 2026-08-06
Feature/gvk contract

### LINKED ISSUE #220: Update pool contract
We need to update the pool Soroban smart contract to accommodate the global view key support:

- Update `newCommitment` event to reflect new data. ( related to #219 )
- Study how to manage different versions of events (new encrypted memos) on the same pool contract (or manage separate smart contracts).
- Store immutable admin view key.
- Expose the public method to allow users to gather the key
- Update `contract-state.rs` to reflect and expose the new functionality.

Fantoni0:
Adds support for GVK-pool contracts. Basically, enables deploying a new type of pool with support for a global view key that:
- Allows the admin to establish his encryption key D at the contract level.
- Introduces a modified NewCommitmentEvent that also encrypts the note secrets under D.
- Introduces a modified NewNullifierEvent that also relates (only to be seen by the pool admin)  a spent event, with the note being spent. For when the GVK support is set to traceable mode.

- We also update the state layer to reflect these changes.

A couple of discussion points. Open to suggestions:
- For now, `events_parsers.rs` drops the new event types. I have tried to make all the changes not to modify current/vanilla pool deployments. Will tackle in a different PR once we have the GVK functionality working and part of deployments.json
- Contract organization: I thought of inherting from original `pool.rs`. But Soroban's 
  `#[contractimpl]` emits `#[no_mangle] extern "C"` symbols the linker never                                                                                                                                                   
  dead-code-eliminate. Which causes different pools (vanilla vs. gvk-pool) to conflict on the methods with same name. I resorted to having 3 pool folders:
  - contracts/pool-core: **NON**-contract crate with the common functionality both pools use: merkletree,  helpers etc. Both pool                                                                                                                                      
  crates depend on it and re-export under the original paths
  - contracts/pool: For vanilla pool with no gvk.
  - contracts/pool-gvk: New gvk pool

 TODO (Probably on  diff PRs):
 - [ ] Update the deployment script to differentiate and instantiate gvk pools when signaled.
 - [ ] Update `events_parsers.rs`
 - [ ] Check if we need to update anything related to Bootnode.
 - [ ] Abstract verifiers  #233 
 
 Closes #220 #280


contracts/pool-gvk/src/pool_gvk.rs
```diff
@@ -0,0 +1,803 @@
...
+        proof: Proof,
+        ext_data: ExtData,
+        sender: Address,
+    ) -> Result<(), Error> {
+        sender.require_auth();
+        let token = Self::get_token(env)?;
+        let token_client = TokenClient::new(env, &token);
+        let zero = I256::from_i32(env, 0);
+
+        if ext_data.ext_amount > zero {
```
cupicmarko, line 622: All tests set `ext_amount` to zero and this path is never tested.

contracts/pool-gvk/src/pool_gvk.rs
```diff
@@ -0,0 +1,803 @@
...
+            .publish(env);
+        }
+
+        // 7. Process withdrawal if ext_amount < 0
+        let token = Self::get_token(env)?;
+        let token_client = TokenClient::new(env, &token);
+        let this = env.current_contract_address();
+        let zero = I256::from_i32(env, 0);
+
+        if ext_data.ext_amount < zero {
```
cupicmarko, line 728: Similar to https://github.com/NethermindEth/stellar-private-payments/pull/461/changes#r3749431503

contracts/pool-gvk/src/pool_gvk.rs
```diff
@@ -0,0 +1,803 @@
...
+    /// new commitments into the Merkle tree.
+    fn internal_transact(env: &Env, proof: Proof, ext_data: ExtData) -> Result<(), Error> {
+        // 1. Merkle root check
+        if !MerkleTreeWithHistory::is_known_root(env, &proof.root)? {
+            return Err(Error::UnknownRoot);
+        }
+        // 2. Nullifier checks (prevent double-spending)
+        for n in proof.input_nullifiers.iter() {
+            if Self::is_spent(env, &n)? {
+                return Err(Error::AlreadySpentNullifier);
```
cupicmarko, line 647: We should add a test that will assert that `AlreadySpentNullifier` by replaying with reused nullifier?

contracts/pool-gvk/src/pool_gvk.rs
```diff
@@ -0,0 +1,803 @@
...
+
+    /// Convert a U256 into a 32-byte big-endian field element.
+    fn u256_to_bytes(env: &Env, v: &U256) -> BytesN<32> {
+        let mut buf = [0u8; 32];
+        v.to_be_bytes().copy_into_slice(&mut buf);
+        BytesN::from_array(env, &buf)
+    }
+
+    /// Maximum absolute external amount allowed (2^248).
+    fn max_ext_amount(env: &Env) -> U256 {
```
cupicmarko: This function, and some others are private functions, so they actually don't have to be duplicate, `#[contractimpl]` only export `pub fn` and only they get `#[no_mangle]`, private functions are simply invisible to the host.

contracts/pool-gvk/src/pool_gvk.rs
```diff
@@ -0,0 +1,803 @@
...
+        }
+    }
+}
+
+/// Storage keys for contract persistent data.
+///
+/// Everything `pool` stores, plus the immutable `AdminViewKey` and `GvkMode`.
+#[contracttype]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) enum DataKey {
```
cupicmarko, line 95: I would recommend moving `DataKey` to core, like this. This is not affected by `#[contractimpl]`, and these two extra keys `AdminViewKey` and `GvkMode` actually won't affect default pool because they're never initialized in storage.

contracts/pool-gvk/src/pool_gvk.rs
```diff
@@ -0,0 +1,803 @@
...
+// `pool`.
+pub use pool_core::{ExtData, hash_ext_data};
+
+/// Contract error types for the GVK privacy pool.
+///
+/// Duplicated 1:1 from `pool::Error` plus the two GVK-specific variants.
+#[contracterror]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum Error {
```
cupicmarko, line 40: Similar to https://github.com/NethermindEth/stellar-private-payments/pull/461/changes#r3749731165

These are always passed as their numerical values. I think that having one error type with all errors incuding GVK specific ones is harmless, because pool will never use them.

cupicmarko (dismissed):
Nice work!

Comments are mostly about tests (can be added later) and duplicate code (can be addressed later, but I think it's better to do it in this PR)

contracts/pool-gvk/src/pool_gvk.rs
Fantoni0: Added tests with positive/negative ext_amount

contracts/pool-gvk/src/pool_gvk.rs
Fantoni0: Same as above

contracts/pool-gvk/src/pool_gvk.rs
Fantoni0: Done

contracts/pool-gvk/src/pool_gvk.rs
Fantoni0: Moved away to pool_core

contracts/pool-gvk/src/pool_gvk.rs
Fantoni0: You are right about `#[contractimpl]`, but I would still prefer to keep them separate. If in the future we extend any of the pools, we do not need to worry about if it affects the other one or not. Duplication is small and I think it's worth the peace of mind

contracts/pool-gvk/src/pool_gvk.rs
Fantoni0: I think it's not the same as the Datakey, because it's public and not public (crate). But as above, I think keeping them separate might be cleaner to avoid confusion.

v0-e (approved):
lgtm!

PR #464, opened 2026-08-07
Updates circom-witness-rs dependency to crates.io version

Fantoni0:
Updates circom-witness-rs
They just [merged](https://github.com/philsippl/circom-witness-rs/pull/36) the changes to make it compatible with ark libs v0.6 
An extra step towards publishing SPP Rust SDK in crates.io

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,6 +190,7 @@ checksum = "6bc66f96ebe2a17a499475b4f94791d379817592ef494171586967ffdc6f95db"
 dependencies = [
  "ark-ec 0.6.0",
  "ark-ff 0.6.0",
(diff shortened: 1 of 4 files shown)


cupicmarko (dismissed):
LGTM!

Nice that we can replace rev pins to specific version from crates.io

Fantoni0:
Ran the `make graph-witness` command to ensure the R1CS remained the same. But I found an issue with how we were patching circomlib. Basically, it was always failing for me, because circom-witness-rs is a build dependency of build.rs, and ran before circomlib, it was patched
Let me know what you think @cupicmarko @v0-e

v0-e (approved):
lgtm!

PR #465, opened 2026-08-09
fix: reject missing and short inputs during witness building

### LINKED ISSUE #462: Strengthen defensive checks
After #434, we changed how we build inputs and how we compile circuits.
For now, missing or short inputs are padded with zeroes and get accepted in the proofs. The proofs will fail later, but it would be nice to reject them at the [witness-building](https://github.com/NethermindEth/stellar-private-payments/blob/main/sdk/witness/src/lib.rs#L128-L172) phase.

Y0sefTamer:
Description
This PR introduces a fail-fast mechanism in compute_witness to prevent missing or short inputs from being zero-padded and accepted during the witness-building phase, which previously led to late proof failures.

Changes Made:

Added a validation check comparing inputs.len() to self.input_hashes.len(). If it's strictly less, it bails with a "missing circuit input signal(s)" error.

Updated the length check for individual input arrays from values.len() > expected_size to values.len() != expected_size to strictly enforce the declared size and reject short inputs.

Related Issue
Closes #462

### DIFF
--- a/sdk/witness/src/lib.rs
+++ b/sdk/witness/src/lib.rs
@@ -131,36 +131,50 @@ impl WitnessCalculator {
             tracing::debug!(elapsed_ms = elapsed_ms(start), error = %e, "calc failed during json parsing");
             e
         })?;
+        
         let mut unknown: Vec<&str> = inputs
             .keys()
             .filter(|key| !self.input_hashes.contains(&fnv1a(key)))
             .map(String::as_str)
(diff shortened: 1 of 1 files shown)


Fantoni0:
Thanks for the contribution @Y0sefTamer :raised_hands: 
Some of the jobs seem to be failing, remember you can always run `./.githooks/pre-push` to check for any issues locally

PR #466, opened 2026-08-09
test(e2e): cover graph.bin witness -> Groth16 prove/verify

### LINKED ISSUE #451: Add test coverage for graph.bin witness -> Groth16 prove/verify
Current tests are based on WASM/er, but with the move to pre-compiled witness graphs (#434) we need to add e2e test coverage for the new design.

RajeshRk18:
Since #434 the CLI, the SDK and the browser compute the witness from a committed `*.graph.bin` operation graph, but the e2e tests still used ark-circom and the circuit WASM. Nothing covered the shipped pipeline.

- Moved the pool e2e tests onto the production path: `generate_proof` now feeds the committed graph through `witness::WitnessCalculator` and the `prover` crate
- Split `generate_proof` into `build_policy_inputs` + `prove_with_graph`, and extracted the contract state sync into `sync_contract_state`.
- Added `Inputs::to_witness_json`, which renders the shared input set as the flat JSON the graph evaluator accepts.
- Added e2e test cases for prove and verify path

Closes #451

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1200,6 +1200,7 @@ dependencies = [
  "parser",
  "program_structure",
  "regex",
+ "serde_json",
  "type_analysis",
  "types",
  "wast",
@@ -2276,6 +2277,7 @@ dependencies = [
  "asp-non-membership",
  "circom-groth16-verifier",
  "circuits",
+ "contract-types",
  "num-bigint",
  "pool",
  "prover",
@@ -2283,6 +2285,7 @@ dependencies = [
  "soroban-utils",
  "tx-planner",
  "types",
+ "witness",
  "zkhash",
 ]
 
--- a/circuits/Cargo.toml
+++ b/circuits/Cargo.toml
@@ -19,6 +19,7 @@ std = [
     "ark-std",
     "num-bigint",
     "num-integer",
+    "serde_json",
 ]
 # Enables circom-witness-rs C++ graph generation (`make witness-graphs`).
 regen-graph = ["dep:circom-witness-rs"]
@@ -40,6 +41,7 @@ ark-snark = { workspace = true, optional = true }
 ark-std = { workspace = true, optional = true }
 num-bigint = { workspace = true, optional = true }
 num-integer = { workspace = true, optional = true }
+serde_json = { workspace = true, default-features = false, features = ["alloc", "std"], optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 # external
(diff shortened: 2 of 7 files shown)


e2e-tests/src/tests/e2e_pool_graph_witness.rs
```diff
@@ -0,0 +1,338 @@
+//! End-to-end coverage for the pre-compiled witness graph path.
+//!
+//! Since #434 the production clients no longer run the circuit WASM. They read
+//! a committed `*.graph.bin` operation graph, compute the witness with
+//! `witness::WitnessCalculator`, and prove with the `prover` crate. These tests
+//! exercise that pipeline:
```
Fantoni0: Using AI for comments is fine, but let's try not to mention Issues/PRs by number. 
Also, no need to give the backward context in comments. Example:

> E2E tests for pre-compiled witness graph path:

better than:

> We used to do ABC, but since issue 123 and PR 345, we now do XYZ. For this reason, we change ... and ...

As the code grows, we will change how we do some things, which is worth discussing (in PRs, issues), but it shouldn't be part of the code itself. 

All good, in general just let's try to simplify a bit.

e2e-tests/src/tests/e2e_pool_graph_witness.rs
```diff
@@ -0,0 +1,338 @@
...
+    );
+    fixture.proof.output_commitment0 = scalar_to_u256(&fixture.env, tampered);
+
+    let outcome = fixture.transact();
+    assert!(
+        matches!(outcome, Err(Err(InvokeError::Contract(code))) if code == Groth16Error::InvalidProof as u32),
+        "expected the verifier to reject a tampered output commitment, got {outcome:?}"
+    );
+    Ok(())
+}
```
Fantoni0: This is basically testing the same as `e2e_pool_2_in_2_out.rs`. I wonder if it would make sense to fold these 2 files together as there's a bit of overlap.
Additionally, if the other 2 tests are already using the new witness building based on graphs, the naming could be confusing. What do you think?

Fantoni0 (changes requested):
Thanks so much @RajeshRk18 🚀 
I like the refactor you did in e2e-tests, I think it's a nice step in the right direction.
I just think that some things can be simplified/merged together.

e2e-tests/src/tests/e2e_pool_graph_witness.rs
RajeshRk18: Yeah, will change it

e2e-tests/src/tests/e2e_pool_graph_witness.rs
RajeshRk18: yepp! `TransactFixture` and test cases in `e2e_pool_graph_witness.rs` moved  to `e2e_pool_2_in_2_out.rs`

Fantoni0 (approved):
Thanks @RajeshRk18, but we require commits to be signed

Issue #386, opened 2026-07-14
UI show transfer/withdraw cost


v0-e:
Use WASM `PrivatePool.estimate(amount)` to show in the transfer/withdraw tabs the number of transactions required to transfer/withdraw the input amount, if the number of transactions is >1. Optionally a confirmation could also be required after clicking the main button to execute the operation.
Can be updated whenever the input is changed by the user.

Patrick-Ehimen:
@v0-e can i be assigned this?

v0-e:
Hey @Patrick-Ehimen, thank you :+1:

PR #467, opened 2026-08-10
feat(app): live transaction-count hint on transfer/withdraw amounts

Patrick-Ehimen:
## 🚀 What’s this PR do?

Shows how many on-chain transactions an amount needs **inline, as the user types**, in the transfer and withdraw tabs only when that number is greater than one.

#436 already shows this figure in the confirmation dialog, i.e. after the user commits. This adds the part #386 describes as "updated whenever the input is changed by the user", so a split is visible while the amount can still be adjusted.

---

### 📎 Related issues

Refs #386 not `Closes`, since #436 may already satisfy it in your view. Happy to close this if so.

---

## 🧠 Context

Reuses the existing `PrivatePool.estimate(amount)` call, so the inline hint and the dialog can't disagree.

- Debounced 300ms otherwise one estimate per digit.
- Only shown when `txCount > 1`, per the issue.
- Out-of-order responses discarded by sequence number; a slow estimate must not describe an amount already changed.
- Recomputed on `pool:selected` (estimate is pool-specific) and `wallet:ready` (an amount typed before connecting is skipped for want of a wallet).
- Cleared on successful submit the field is reset programmatically and fires no `input` event.
- Best-effort like `txCountRow`: a failed estimate clears the hint rather than raising.

Note: `txCountRow()` guards with `if (!txCount)`, so the dialog still renders "1 transaction" for single-transaction ops, which #386 asks to suppress. Left untouched as out of scope say the word and I'll align it.

---

## ✅ Required Checklist

- [ ] I’ve tested this locally, run .githooks/pre-push and provided test instructions for maintainers if needed
- [ ] I’ve updated or created tests if needed

### DIFF
--- a/app/js/ui/transactions.js
+++ b/app/js/ui/transactions.js
@@ -97,21 +97,29 @@ function onEnterUnlessBusy(input, button, handler) {
     });
 }
 
-// Builds a confirmation-dialog row with the number of transactions the action
-// requires. Best-effort only: a failed estimate must not block the action.
-async function txCountRow(amountValue) {
+// Best-effort estimate of how many on-chain transactions an amount requires.
+// Returns null on any failure — the confirmation dialog and submit path already
+// surface anything that truly blocks the operation.
+async function estimateTxCount(amountValue) {
     try {
         const session = await ensureAppPool();
         const estimate = await session.estimate(amountValue);
-        const txCount = estimate?.txCount;
-        if (!txCount) return null;
-        return {
-            label: 'Transactions',
-            value: `${txCount} transaction${txCount === 1 ? '' : 's'}`,
-        };
+        const txCount = Number(estimate?.txCount ?? 0);
+        return txCount > 0 ? txCount : null;
     } catch {
         return null;
     }
+}
+
+// Builds a confirmation-dialog row with the number of transactions the action
+// requires. Best-effort only: a failed estimate must not block the action.
+async function txCountRow(amountValue) {
+    const txCount = await estimateTxCount(amountValue);
+    if (!txCount) return null;
+    return {
+        label: 'Transactions',
+        value: `${txCount} transaction${txCount === 1 ? '' : 's'}`,
+    };
 }
 
 // Re-estimate callbacks for every bound amount input, so a pool change can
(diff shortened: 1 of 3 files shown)


app/js/ui/transactions.js
```diff
@@ -349,6 +412,7 @@ export const Transactions = {
         App.events.addEventListener('pool:config', updateMoveFundsBalance);
         App.events.addEventListener('pool:selected', updateMoveFundsBalance);
         App.events.addEventListener('balances:updated', updateMoveFundsBalance);
+        App.events.addEventListener('pool:selected', refreshCostHints);
```
Fantoni0, line 415: This listener and the regular page setup might conflict when setting the pool session. Especially if the user hops across multiple modes (e.g. transfer and withdraw).

app/js/ui/transactions.js
```diff
@@ -111,6 +114,61 @@ async function txCountRow(amountValue) {
...
+        if (!amount.ok || amount.value <= 0n || !App.state.wallet.address || !isRuntimeReady()) {
+            hint.textContent = '';
+            return;
+        }
+        timer = setTimeout(async () => {
+            try {
+                const session = await ensureAppPool();
+                const estimate = await session.estimate(amount.value);
+                if (current !== seq) return;
+                const txCount = Number(estimate?.txCount ?? 0);
```
Fantoni0, line 152: This is mostly duplicated from tcCountRow()

Fantoni0 (changes requested):
Thanks for the contribution @Patrick-Ehimen 💯 
I like the idea a lot, but I think some race conditions might happen.
My advice would be, rather than showing the number of steps a hint when configuring the transaction, you could show it in the confirm (app/js/confirm) modal. They already have access to the `txCountRow()` function, and you don't need to worry about calling ensureAppPool and avoid racing issues

Fantoni0 (changes requested):
Thanks @Patrick-Ehimen 
I was checking the deployed version on main, and noticed we already show a transaction-count hint on the confirm modal:

Image

So not sure this PR would be adding anything else on this aspect :( 
That being said, I think your changes for hardening the pool session are good and can still be merged. If you could please:
- Remove the html hints.
- Any logic related to transaction count/hints that is no longer needed
- Take a look to `navigation.js`, I think it could also use the pendingSession guard (on the disconect function we call closeAppPool but a createAppPool might already be running)

I think we can still merge this 🙌

Fantoni0:
Closed as stale

PR #468, opened 2026-08-10
refactor(circuits): replace build.rs with a dedicated circuit-builder binary

### LINKED ISSUE #463: Move build logic
As @v0-e suggested:

> moving circuits/build.rs logic into its own binary. Circuit artifacts are such a central part of our process and safety/security, and a dedicated executable could be a more idiomatic design.

Right now circuits/build.rs is a lib crate with a build.rs script. If it were made a binary crate with a wrapper bash script, we could trigger this behavior in other parts of the codebase, and not only when the circuits crate is compiled. 
An example could be publishing compiled artifacts in CI rather than having them be part of the committed codebase.

Not urgent. But it would be nice to have.

RajeshRk18:
## 🚀 What’s this PR do?

Closes #463 

- Replaces `circuits/build.rs` with `tools/circuit-builder`, a binary with `vendor`, `patch`, `compile`, `keys`, `snapshot` and `verify` subcommands. 

- CI builds the artifacts once in a reusable workflow and shares them between jobs. Nine workflows each ran `cargo build -p circuits` before.

- Two manifests are new. `target/circuits-artifacts/manifest.json` records each circuit's r1cs and wasm digests, its `.circom` source closure, and the toolchain. `deployments/testnet/circuit_keys/keys.manifest.json` binds each committed key pair to the sha256 of the R1CS its setup consumed. `circuit-builder verify` checks both on every PR, so a circuit edit that needs a key rotation fails review instead of surfacing after a deploy.

- The GPL-3.0 circom dependencies move with the code. `circuits/LICENSE`, `README.md` and `deny.toml` now name `tools/circuit-builder` instead of `build.rs`.

- Some existing bug fixes such as incorrect regex, bare includes not resolved, and not-good dep tracking.
---

### 📎 Related issues

#123

---

## 🧠 Context

- The profile level is gone from target/circuits-artifacts/. sdk/web/scripts/build.sh already documented these artifacts as profile-independent.

- e2e tests now load the committed testnet keys. They previously read testdata/ which build.rs filled with a fresh random trusted setup on every CI run. They now prove and verify with the deployed pair.

Unrelated to this PR: sdk/web's telemetry tests flake about 1 run in 20 on main via a shared RING_BUFFER and a poisoned TEST_MUTEX. It tripped during pre-push here.

---

## ✅ Required Checklist

- [x] I’ve mentioned this PR in the project LinkedIn group (required for review so maintainers can see a human behind the contribution).


Fantoni0:
Hi @RajeshRk18, thanks for taking this on! The direction here is good: fixing some issues, breaking up the build.rs script, binding keys to R1CS etc. And I don't want that work to feel wasted. That said, I'm not going to merge this PR as-is, for a few reasons:                                                                                                                                 
  - #463 wasn't labeled as contributor-friendly. When you proposed to take a look, I expected some minor changes. But a change of this size (63 files, many lines, touching CI, licensing config, key materials,  and SDK/CLI paths) is a lot of surface area for us to take on trust from third-party contributions.
  - It modifies licensing
  - I did a preliminary review and found some issues that will require some iteration: circuit-builder can overwrite the ceremony label on untouched keys, and there is some leftover code from before the refactor.
  
  Given the scope, I'd rather close this than ask for a large revision cycle. If you'd like to continue on this, you can split some parts into smaller and dedicated PR, e.g., fixing some of the issues you mentioned.

PR #471, opened 2026-08-12
Add pre-sign wasm-bindgen e2e tests and CI

cupicmarko:
## Summary

Adds the pre-signing browser e2e suite for the web client and the CI workflow that
runs it. The tests drive deposit, transfer and withdraw end-to-end against testnet
in headless Chrome via `wasm-bindgen-test`, stopping at the transaction-signing
boundary.

## What the tests assert

Each flow halts at the signing boundary and asserts **status `"failed"` +
SEP-0043 code `-4` + empty transaction hashes + having reached the `"sign"`
progress stage**. All four together, deliberately: a genuine upstream failure
cannot masquerade as a green test, because it would not reach the `sign` stage
with exactly code `-4`.

A dual-mode signer backs this. Sentinel mode returns `-4` from `signTransaction`
and `signAuthEntry` for the flows under test; a real `LocalSigner` signs only the
setup deposits that seed on-chain notes for transfer and withdraw. Those setup
transactions are genuinely submitted — the "stop before signing" scope applies to
the flows under assertion, not to fixture setup.

The six e2e tests are `#[ignore]d` by default, so the existing PR-time `wasm-test`
job keeps running this crate's other tests without secrets or a static server.
Running them requires `-- --include-ignored`.

## Files

**Test suite and supporting sdk/web changes**
- `sdk/web/src/client/e2e_tests.rs` — the six tests, the dual-mode signer, and
  blob-wrapped module-worker loading (the wasm-bindgen runner serves its page from
  a per-run tempdir, so cross-origin module workers never start; loaders are
  fetched from the CORS static server, rewritten to absolute URLs and wrapped in
  `blob:` URLs — gloo_worker's own pattern)
- `sdk/web/src/signer.rs` — sentinel signer, plus spike tests pinning the mechanism
- `sdk/web/src/client/execute/mod.rs` — error-mapping spike test for the rejection path
- `sdk/web/src/client/mod.rs` — module wiring under `cfg(test)`
- `sdk/web/build.rs` — `rerun-if-env-changed` for the `E2E_*` vars, so `option_env!`
  config cannot go stale
- `sdk/web/src/circuits.rs` — **bug fix**: the `window.fetch` shim was never
  restored, which panicked every later RPC-touching test in the same page. Now
  captured and restored via a `Drop` guard.

**Tooling and CI**
- `sdk/web/scripts/e2e-browser-test.sh` — owns the run lifecycle: exports the
  account env file, builds `sdk/web/dist` when missing, serves it with CORS on
  `E2E_STATIC_ORIGIN`, polls for readiness, resolves `CHROMEDRIVER`, and raises
  `WASM_BINDGEN_TEST_TIMEOUT` to 600s
- `deployments/scripts/e2e-accounts-setup.sh` — idempotent provisioning of the two
  accounts this suite drives: A as the account under test, B as the registered
  transfer recipient. Secrets go only to a git-ignored, mode-600 env file.
- `.github/workflows/e2e-webclient.yml` — runs on push and PR to `main`
- `sdk/web/tests/README.md` — prerequisites, provisioning, running
- `.gitignore` — ignores the generated account material

## Running it

```bash
cargo build -p circuits && cargo build -p circuits --release
deployments/scripts/e2e-accounts-setup.sh    # once; needs the spp CLI on PATH

sdk/web/scripts/e2e-browser-test.sh \
  cargo test --target wasm32-unknown-unknown -p stellar-private-payments-sdk-web \
  -- --include-ignored

### CODE REVIEW

File: .github/workflows/e2e-webclient.yml [hunk 1]
```diff
@@ -0,0 +1,222 @@
+name: E2E Web Client
+
+# PR #471 broke because the CI secret E2E_POOL_CONTRACT held a pool address
+# from a superseded deployment generation, and the workflow's own fallback
+# default baked the SAME stale address into the YAML as a literal — two
+# independent copies of a contract address, disconnected from
+# deployments.json, both wrong in the same way, with nothing to notice the
+# drift.
+#
```
Fantoni0: Using AI for documentation is quite useful, but I personally find it too verbose.
I particularly don't like mentioning PRs or issue numbers. I feel like something from those discussions is worth mentioning in the comment. It should be directly addressed/copied instead of sending the reader to other places. Especially if it's already a 4-paragraph comment. 
In general, the comments read as "justification" from the AI model rather than providing concise information.

There are multiple instances of this in the PR. Some are fine, some are a little more verbose. But it's subjective, and there's a tradeoff.

sdk/web/build.rs
```diff
@@ -71,6 +71,22 @@ fn main() {
...
+    for var in [
+        "E2E_STATIC_ORIGIN",
+        "E2E_RPC_URL",
+        "E2E_POOL_CONTRACT",
+        "E2E_ACCOUNT_A_ADDRESS",
+        "E2E_ACCOUNT_A_SECRET",
+        "E2E_ACCOUNT_B_ADDRESS",
+        "E2E_ACCOUNT_B_SECRET",
+    ] {
+        println!("cargo:rerun-if-env-changed={var}");
```
Fantoni0, line 83: This will leave information about the secrets in ` target/debug/.fingerprint/<pkg>/run-build-script-build-script-build.json`.
I think it's mostly fine, because these are test accounts. But worth mentioning.

Fantoni0 (dismissed):
Looks good to me. I just left some minor comments, most of them style related. So up to you. Feel free to ignore.

v0-e (dismissed):
lgtm also, but I agree with @Fantoni0 on the comment verbosity. Also the PR stack is introducing 9k lines so perhaps there is some room for simplification after we get this test design stabilized and working :+1:

cupicmarko:
@Fantoni0 @v0-e Can you review [this commit specifically](https://github.com/NethermindEth/stellar-private-payments/pull/471/changes/685c7f7499970d884f912667192ced701f9693e6)? It fixed a bug that I found in CI, it should be merged with tests

v0-e (approved):
The commit looks good, I'm unsure how much impact blocking on it will have on other storage calls as the deleted comment suggests, but we can try it :+1:

Fantoni0 (approved):
LGTM!
Much cleaner now. Let's see the impact of the storage modifications 🤞

PR #473, opened 2026-08-12
Implement end-to-end browser tests

cupicmarko:
## Summary

Adds the real-Freighter browser e2e suite (11 tests), a shared preflight used by
both browser suites, local tooling to set up / repair / prune the e2e
environment, and a correctness fix in the SDK's spent-nullifier lookup that the
suite exercises end to end.

Stacks on `tests/e2e-webclient-presign` — that branch merges first.

## Spent-nullifier fix (`sdk/stellar/src/contract_state.rs`)

`is_nullifier_spent` made a single unpaginated `getEvents` call anchored at the
pool's `deployment_ledger`. `getEvents` scans a bounded ledger window per call
and returns an **empty page, not an error**, when the match lies beyond it.

Measured against the live RPC: a nullifier spent ~30,000 ledgers after the
deployment ledger returned zero events, while a query starting within ~10,000
ledgers of the spend found it immediately. Both ranges sit well inside the RPC's
retention window, so this is not `oldestLedger` pruning.

The consequence is a false negative in a privacy-critical path: a note spent
well after deployment reported **unspent**, so disclosure receipts for
already-spent notes verified as fully valid. It worsens as a pool ages, since
`deployment_ledger` only grows staler relative to the spend.

The lookup now paginates by cursor (bounded at 64 rounds) until it matches or
the cursor stops advancing. It also collapses two speculative topic-name casings
(`new_nullifier_event` / `NewNullifierEvent`) into the one the contract actually
emits.

**The topic filter is unchanged** — two segments, `[event_name, nullifier]`,
before and after. A new regression test pins this by decoding a real on-chain
`NewNullifierEvent` from XDR and asserting byte-for-byte equality segment by
segment; forcing it to one segment fails outright. The code comment records that
an earlier attempt concluded single-segment by reading the `#[contractevent]`
macro's `if` branch without the `else` — worth knowing, because that conclusion
is already in this repo's history where a reviewer will meet it.

## The suite (`e2e-freighter/`)

Playwright against a locally served app and live testnet. Each test starts from
a provisioned Chrome profile with the real extension loaded, and every signature
is approved through Freighter's own UI — no mocked provider, no injected key.

| # | Test | Exercises |
|---|------|-----------|
| 01 | connect | wallet connect, network and address |
| 02 | deposit | deposit confirmed SUCCESS on-chain by tx hash |
| 03 | rejection | reject at the Freighter prompt, assert clean recovery |
| 04 | deposit-withdraw | full round trip |
| 05 | deposit-transfer | transfer to a registered recipient |
| 06 | disclose-basic | receipt verifies VALID / UNSPENT |
| 07 | disclose-spent | spent note reads SPENT while an unspent note still verifies fully valid |
| 08 | disclose-lifecycle | multi-note lifecycle across two spends, re-verified against changed chain state |
| 09 | disclose-negative | garbage input, tampered proof, wrong context |
| 10 | advanced-transfers | transfer through the Advanced composer |
| 11 | failure-modes | over-withdraw, over-transfer, unregistered recipient, deposit above the on-chain cap |

Tests 06–08 are what make the nullifier fix observable from outside the SDK: 07
fails on the unpaginated lookup and passes on the paginated one.

`scripts/provision.sh` builds and snapshots the profile. Two things there are
load-bearing and commented as such — the snapshot **must keep** the extension's
MV3 service-worker script cache (without it a restored profile answers
"Receiving end does not exist" and the app reports Freighter as not detected,
with the wallet intact in storage the whole time), and it **must drop** Chrome's
per-machine extension caches and singleton locks.

## Shared preflight (`scripts/e2e-preflight.sh`)

Both browser suites now run one preflight that fails with a specific remedy
rather than a downstream symptom: toolchain, account env file (present, mode,
gitignored, vars, address formats), pool-vs-`deployments.json`, RPC
reachability, accounts funded and registered, circuit artifacts, vendored
extension pin, profile snapshot integrity.

The pool-vs-deployments check earns its place: `sdk/web` and the CLI both embed
`deployments.json` at **compile time** via `include_str!`, so a stale binary
silently tests a different deployment than the one on disk.

On CI the env-file checks skip (secrets are injected instead) while
`env.vars.required` still verifies against the environment for real.

## Local tooling

| Command | Does |
|---|---|
| `make freighter-smoke` | three-test subset; sets up what's missing, serves, runs, stops the server |
| `make freighter-e2e` | same, whole suite |
| `make freighter-repair` | salvage after a redeploy: rebuild `spp`, re-register accounts, prune dead state, re-verify |
| `scripts/e2e-cleanup.sh` | prune wallet-DB state for contract generations no `deployments.json` names |

The run targets stop the server on pass, fail or Ctrl-C, killing the process
group — `make` execs trunk as a child, so killing `make` alone orphans it
holding the port. A server already on the port is reused and left running.

`e2e-accounts-setup.sh` gains `--reregister`. A redeploy mints a new public-key
registry, so accounts keep their keypairs and funding but lose their
registration; previously `--verify` said "run without --verify", which verified
again and died identically, leaving only `--force` — which regenerates all four
keypairs and abandons funded accounts to fix a registration.

## CI

Smoke subset on PRs, whole suite on demand from the Actions tab. **Both build
this checkout and serve it on `localhost:8000`.** Previously the PR-time run
pointed at the deployed GitHub Pages app and skipped every build step, so a PR
was tested against whatever happened to be deployed rather than its own code.

With both levels building identically, the smoke/full matrix and its six
per-step conditionals collapse into a single job whose only difference is which
files run. A concurrency group serializes runs: the suite drives four shared
testnet accounts against one pool, so overlapping runs spend each other's notes
and fail in ways that look like product bugs.

## Verification

All eleven tests pass locally against testnet, including the on-chain
confirmation of the nullifier fix (07: spent note reads SPENT, unspent note
still fully valid).

CI has not yet run this workflow green — the local runs used a testnet
deployment whose `deployments.json` is deliberately not included here, so a
first CI run may hit the RPC sync gap until the deployment is refreshed.

## Review notes

- The job name is now a static `Freighter e2e`; branch protection pinned to the
  old dynamic names needs updating or PRs will sit unreported.
- The suite spends real testnet XLM; a full run takes ~20 minutes.


sdk/stellar/src/contract_state.rs
```diff
@@ -635,31 +660,52 @@ impl StateFetcher {
...
+        // only grows staler over a pool's lifetime — silently produces a
+        // false "unspent" for any note spent more than one scan window
+        // after deployment. Keep paginating via the returned `cursor` (the
+        // documented continuation mechanism) until a match is found or the
+        // cursor stops advancing (caught up to the events tip).
+        const MAX_ROUNDS: usize = 64;
+        let mut start = EventStart::Ledger(pool.deployment_ledger);
+        let mut last_cursor: Option<String> = None;
+
+        for _ in 0..MAX_ROUNDS {
```
v0-e: Instead of this, we never got around to around to using simulate/`is_spent` for the nullifier check?
It should be usable now (we made it pub in the contract).
Else, I think this will be limited by the 7-day retention period of the RPC. `MAX_ROUNDS` = 64 at max 10k ledger per request would also limit this approach at ~1 month of history.

deployments/testnet/deployments.json
v0-e, line 1: If redeployment is not required, lets see if the bootnode can hold the previous deployment 😄

Fantoni0:
I was able to run the tests. Some comments:

- `e2e-freighter/scripts/setup.sh` was not able to handle the bootnode required modal. I had to manually accept it, otherwise it timed out
image

- Bootnode issues aside, I had to recompile the SPP CLI tool to gather the latest deployment (Maybe it could be a signal in the pre-flight)

- `bash scripts/e2e-preflight.sh` complained about a server not being run in localhost (it was running). But ` e2e-freighter/scripts/run-all.sh` did pass the check.

- Some tests failed on my machine, but it could be related to timeouts. So if they pass the CI. It would be fine by me.

cupicmarko:
@Fantoni0, do you maybe still have logs so I could debug?

> Some tests failed on my machine, but it could be related to timeouts. So if they pass the CI. It would be fine by me.

cupicmarko:
Redeployment was intentional, we reached full capacity of the Merkle tree.

Fantoni0 (approved):
Great work @cupicmarko 
Was a tough one, but you did it. Now we have fully working E2E tests

v0-e (approved):
lgtm :)

PR #475, opened 2026-08-14
Move bootnode DB migrations

v0-e:
Moves bootnode database migrations to own file rather than having them in `.rs` src.


Fantoni0 (approved):
LGTM!

PR #476, opened 2026-08-14
Event-based bootnode

v0-e:
Large refactor of the bootnode logic. Instead of us storing the pages fetched from the main RPC, we store now only the events.
While the previous approach was initially simpler since we would only provide the fetched pages as-is to the client, and we would avoid doing our own pagination logic, the approach evolved to a needlessly complex design which required some finicky page management to support some storage efficiency (e.g., page compression).

As we move to event-based storage, now we also support fetching using any ledger as well as event limits on provided pages.

Previous rules apply:
- bootnode provides by default history until 5 days before current date;
- handoff is characterized with a `-32002` code (requests with `startLedger`/cursor more recent than 5 days);
- out-of-sync / bootnode catching up is characterized with a `-32004` code (until bootnode reaches ledger tip).

We now move to a more efficient and maintainable version of the bootnode.


tools/bootnode/src/indexer.rs
```diff
@@ -126,4 +105,14 @@ impl Indexer {
...
+
+    async fn mark_archive_ready(&self) -> anyhow::Result<()> {
+        if self.state.archive_ready.load(Ordering::Relaxed) {
+            return Ok(());
+        }
+        self.state.storage.set_archive_ready().await?;
+        self.state.archive_ready.store(true, Ordering::Relaxed);
+        tracing::info!("archive ready: ingestion crossed retention cutoff");
+        Ok(())
+    }
```
Fantoni0, line 117: Just one question, once we reach the retention cutoff, is this never reset/checked again?
Not sure if this could be an issue for posterior requests or if the indexer stays behind.

v0-e: Indeed, this was a choice to keep simplicity high. Since the cutoff is 5 days, then the bootnode/indexer would have to be out of sync for 5 days which would mean a deeper issue in the network (or bootnode).
Can happen, but is an unlikely edge-case and at least for this version, I think we can review this later.

Fantoni0 (approved):
LGTM!

PR #480, opened 2026-08-17
Fix bug with worker running after closing the page

cupicmarko:
Each page (index.html/admin.html) spawns its own dedicated storage worker, which installs an OPFS SQLite VFS (sahpool). Navigating between them raced the outgoing worker's OPFS handle release against the incoming worker's acquire, and the failure was consistently misreported:

- OpfsSAHError's Display/Debug never interpolate the wrapped JsValue, so the lock-detection heuristic (is_opfs_locked_error) was matching against a string that could never contain the real DOMException text, always falling through to a generic "Failed to initialize" toast.
- After fixing detection to inspect the DomException directly, a retry loop was added to tolerate the async release race — but pause_vfs() refuses to release handles while SQLite still has files open, and the worker never closed its connection, so the "release on pagehide" path was a silent no-op.

Fixes, in order of effect:
- is_opfs_locked_error now matches on OpfsSAHError::CreateSyncAccessHandle and checks the wrapped DomException's name directly.
- storage worker retries sahpool::install for ~2s on a detected lock before surfacing the "another tab" message, tolerating the normal async handle-release delay.
- New Pause request: on the page's `pagehide` event, the main thread fire-and-forgets a Pause call; the worker now drops its live SQLite connection first (closing the VFS file) so pause_vfs() can actually succeed, releasing the OPFS pool before the browser tears the worker down.  

Note: The issue is reproduced in Firefox

### DIFF
--- a/app/js/wasm-facade.js
+++ b/app/js/wasm-facade.js
@@ -126,8 +126,37 @@ export async function ensureStorage() {
     if (!storageHandle) {
         storageHandle = await Storage.open();
         bindAppStorage(storageHandle);
+        installStoragePauseOnUnload();
     }
     return appStorageInstance;
+}
+
+let pauseOnUnloadInstalled = false;
+
+/**
+ * Ask the storage worker to release its OPFS sync access handles as soon as
+ * this page starts unloading, instead of waiting for the worker to be torn
+ * down by the browser (which happens asynchronously and can race the next
+ * page's worker trying to acquire the same handles — surfacing as a false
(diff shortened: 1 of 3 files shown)


v0-e (approved):
lgtm!
We could also explore [Web Locks](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API#concepts_and_usage). Looks like a potential nice solution for this. 
Added #555.

PR #481, opened 2026-08-17
More bootnode tests

v0-e:
Some more tests.


tools/bootnode/tests/common/mod.rs
```diff
@@ -137,6 +137,13 @@ pub fn event_id_for_ledger(ledger: u32) -> String {
     format!("event-{ledger:010}")
 }
```
Fantoni0: Using random_even_id makes `event_id_for_ledger` dead code?

v0-e: Indeed :+1:, a leftover, removed it.

Fantoni0 (approved):
LGTM!

Issue #374, opened 2026-07-09
Add negative tests for malformed public inputs, stale ASP roots, duplicate nullifiers, replay attempts, empty proofs, max deposit edge cases, zero-value edge cases, and tree-capacity boundaries.


Fantoni0:

zexoverz:
Most of this list looks covered already by #485 and #507, which pinned replay, deposit bounds, zeroed proofs, non-canonical inputs, zero-value and the unknown/evicted root cases.

The one that cannot go in `contracts/pool/src/test.rs` is tree capacity through `transact`. `insert_two_leaves` runs at step 9, after `verify_proof`, so with a mock proof the call always stops at `InvalidProof` and never reaches the boundary. That one needs `e2e-tests`, where real Groth16 proofs exist. Happy to take it if nobody has.

Galmanus:
I'd like to take the slice of this that can be tested without the proving harness. Reading `internal_transact`, six of these fail at guards that run before `verify_proof`, so they can be exercised with a crafted `Proof`/`ExtData` and no real ZK proof:

- unknown root -> `UnknownRoot` (check 1)
- spent nullifier / replay -> `AlreadySpentNullifier` (check 2)
- `ext_data_hash` mismatch -> `WrongExtHash` (check 3)
- wrong `public_amount` -> `WrongExtAmount` (check 4)
- non-canonical BN254 public input -> `NonCanonicalPublicInput` (validated at the top of `verify_proof`, before the verifier call)
- empty proof -> `InvalidProof`

I'll send these as unit tests on the pool contract.

Two items are worth scoping explicitly so the coverage claim stays honest:

- Duplicate nullifiers are already constrained in the circuit (`transaction.circom:123-131` forces `sameNullifiers.out === 0` for every pair, alongside value conservation at `:137`). The pool itself does not re-check distinctness, so a true negative test here needs a real proof against the circuit rather than a contract-only fixture. Happy to leave that for the proving-harness path, or to add a defensive contract-level dedupe if you'd want one.
- The tree-capacity boundary needs a filled tree (real insertions/proofs), also better suited to the harness path.

Assigning myself the pre-verify set unless someone is already on it.

zexoverz:
Four of those six are already pinned on main by #485 and #507. `UnknownRoot`, `AlreadySpentNullifier`, `WrongExtAmount` and three `NonCanonicalPublicInput` cases each assert the specific error, not just that it failed.

[...]

PR #485, opened 2026-08-18
test(pool): negative vectors for replay, deposit bounds and proof rejection

zexoverz:
Towards #374.

Five transact-level negative tests. The existing ones all fail before `verify_proof`, so everything past the spent-check was uncovered.

Replay of a spent nullifier, a deposit above the maximum (refused before any token moves, so an unfunded sender cannot mask the check), a deposit exactly at the maximum to pin the bound as inclusive, an all-zero proof, and duplicate nullifiers inside one call.

Two things surfaced that are worth your judgement rather than mine.

These are the first tests to reach `verify_proof`, and the error a caller gets there is misleading. `Groth16Error::MalformedPublicInputs` is 1, the same numeric code as this contract's `Error::NotAuthorized`, so a rejected proof reads at the pool boundary as an authorization failure. I asserted the current behaviour rather than the behaviour I expected, so the test fails if the codes are ever separated.

The duplicate-nullifier test asserts only that the result is *not* `AlreadySpentNullifier`. The spent-check compares each nullifier against stored state and never against the others in the same call, so pairwise distinctness rests entirely on the circuit. I have not called that a defect, since it may well be intentional, but nothing recorded it.

One note on style: the existing tests assert `is_err()`, which also passes on `Abort`. Mine assert the specific error, which is how a missing `mock_all_auths` in an early draft got caught rather than passing silently.

### DIFF
--- a/contracts/pool/src/test.rs
+++ b/contracts/pool/src/test.rs
@@ -1032,3 +1032,163 @@ fn test_pool_events_exact_shapes() {
     let expected_nullifier = NewNullifierEvent { nullifier }.to_xdr(&env, &contract_id);
     assert_eq!(events.events()[1], expected_nullifier);
 }
+
+/// Marks a nullifier as spent directly in pool storage, mirroring what a
+/// successful `transact` does in step 6. Used to reach the replay path without
+/// needing a real Groth16 proof for the first spend.
+fn mark_nullifier_spent(env: &Env, pool_id: &Address, nullifier: &U256) {
+    env.as_contract(pool_id, || {
+        env.storage()
+            .persistent()
+            .set(&crate::pool::DataKey::Nullifier(nullifier.clone()), &());
+    });
+}
+
+/// Replay: a nullifier that is already spent must be refused, and refused for
+/// that reason rather than falling through to proof verification.
+#[test]
+fn transact_rejects_replay_of_spent_nullifier() {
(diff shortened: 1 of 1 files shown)


zexoverz:
Six days, no review yet. Two things in the description want a maintainer call rather than mine.

`Groth16Error::MalformedPublicInputs` is 1, the same numeric code as this contract's `Error::NotAuthorized`, so a rejected proof reads at the pool boundary as an authorization failure. I asserted the current behaviour, so the test fails if those are ever separated. Separately, the duplicate-nullifier case rests entirely on the circuit, since the spent-check compares each nullifier against stored state and never against the others in the same call. I did not call that a defect.

Happy to split either into its own PR if that reviews more easily.

Fantoni0 (approved):
Thanks for your contribution @zexoverz . LGTM! 💯 

Sure, feel free to create some other PRs to address the issues you mention.
We are actively working on the protocol, so we cannot guarantee fast reviews. But we try to review third party contributions within a week.
For future PRs, AI documentation is fine, but we are trying to keep comments shorter, as sometimes it can get verbose

Fantoni0:
Also, please notice we require signed commits :pray:

zexoverz:
Signed and force-pushed, both commits verify now. Noted on the review turnaround, sorry for the early ping @Fantoni0

PR #490, opened 2026-08-19
Dedicated circuit compiler

### LINKED ISSUE #463: Move build logic

v0-e:
Mostly moves `circuits/build.rs` to its own bin at `tools/circuit-compiler`.

Initially I was thinking of putting the bin in `circuits` but the code there is being used as a library for `e2e-tests` and it would be too much of a mix-up of library code and tooling. We can also do that by simply moving that lib code to another crate (e.g., `e2e-tests`).

Also removes circuit profiles handling (release vs debug) since that makes little sense now. Circuits are always compiled in "release" mode (e.g., circom `-O2`), and the `circuit-compiler` itself is an executable.

More features to follow.

Closes #463.


.githooks/pre-push
```diff
@@ -2,7 +2,7 @@
 set -e
 
 echo "⚠️ Some tests are ignored because they take a long time"
-echo "⚠️ you can run them with ('BUILD_TESTS=1 cargo test -p circuits -- --ignored') "
+echo "⚠️ you can run them with ('make circuits TESTS=1 && cargo test -p circuits -- --ignored') "
```
v0-e, line 5: Consider refactoring current the `make` design. Helpful but deserves a facelift, e.g., it's main purpose was `app/` builds, but now is more than that.
Consider also switching over to `just` as it has some pros in most of our cmds, e.g., here on args `make circuits TESTS=1` vs `just circuits --tests`.

tools/ceremony-cli/src/main.rs
Fantoni0: Probably needs to be `make circuits` now

circuits/LICENSE
```diff
@@ -23,4 +23,4 @@ The circuit files in this directory import templates from `circomlib`, which is
 
 * *Compatibility:* Linking *Apache 2.0* source code with an *LGPLv3* library is a legally supported configuration.
 
-See the root `LICENSE` and `circuits/COPYING` files for the full license texts.
\ No newline at end of file
+See the root `LICENSE` and `circuits/COPYING` files for the full license texts.
```
Fantoni0, line 20: Not sure this belongs here anymore.
My understanding is that circuits now simply include the circomlib LGPLv3 dependency. So point 2 makes sense. But Point 1 about GPLv3, should probably move to `tools/circuit-compile/LICENSE` ?

Fantoni0 (changes requested):
Looks pretty clean. Left some comments.
I also think the CI is failing because `e2e-freighter.yml` and probably other helper scripts for the E2E still use the old `cargo build -p circuits`

tools/ceremony-cli/src/main.rs
v0-e: :+1: fixed

circuits/LICENSE
v0-e: True, fixed that :+1:

Fantoni0 (approved):
LGTM!

PR #491, opened 2026-08-20
CLI use configured bootnode

### LINKED ISSUE #488: CLI not connecting to bootnode
The CLI is storing the bootnode URL, but never actually uses it to fetch data.

v0-e:
Fix to actually load the configured bootnode URL and use it.

Fixes #488.


Fantoni0 (approved):
Once CI passes, LGTM!

PR #494, opened 2026-08-20
Refactor/poseidon2

Fantoni0:
Removes the local poseidon2 crate (Horizen). 
We use taceo-poseidon2, which actually uses Horizen internally, but they expose the T sizes we use.


Fantoni0:
CodeQL is failing because the refactor removed the codebase's only Python script. 
Will try to disable Python form the CodeQL list.
I think we can merge and let CodeQL re-detect  the language change in main

v0-e (approved):
Great! 🚀

v0-e:
Yes, guessing CodeQL issue will resolve on main

cupicmarko (approved):

PR #495, opened 2026-08-20
Refactor witness graph generation

v0-e:
Adds a `--graphs` option to the compiler to build the witness graph bins.

Introduces a new binary for graph generation to circumvent the one-graph one-cargo-build of `circom-witness-rs` v0.3. This potentially can be replaced in the future by the new [`circom-witness-graph-builder`](https://github.com/philsippl/circom-witness-rs/tree/master/graph-builder) utility. However that is linked with `circom-witness-rs` v0.4 which currently doesn't compile for wasm32.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -40,7 +40,7 @@ if ! typos; then
     exit 4
 fi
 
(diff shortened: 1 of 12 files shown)


Fantoni0 (approved):
LGTM!

PR #497, opened 2026-08-20
Circuit artifacts bundle

v0-e:
Adds script to create the lock file (circuit versioning), verify it, and pack circuits.
Bundle release is based on the new CI file, published whenever the `circuits.json` version is bumped.

### DIFF
--- a/deployments/scripts/circuit-artifacts.sh
+++ b/deployments/scripts/circuit-artifacts.sh
@@ -149,7 +149,7 @@ write_lock() {
     printf '    "commit": "%s",\n' "$commit"
     printf '    "circom": "%s",\n' "$(trim_file "$ROOT/circuits/circom.lock")"
     printf '    "circomlib": "%s",\n' "$(trim_file "$ROOT/circuits/circomlib.lock")"
-    printf '    "circomWitnessRs": "%s"\n' "$witness"
+    printf '    "circom-witness-rs": "%s"\n' "$witness"
     printf '  }'
     for stem in $STEMS; do
       printf ',\n'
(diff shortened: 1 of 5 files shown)


deployments/scripts/circuit-artifacts.sh
```diff
@@ -0,0 +1,232 @@
...
+    printf '    "repository": "%s",\n' "$GITHUB_REPO"
+    printf '    "commit": "%s",\n' "$commit"
+    printf '    "circom": "%s",\n' "$(trim_file "$ROOT/circuits/circom.lock")"
+    printf '    "circomlib": "%s",\n' "$(trim_file "$ROOT/circuits/circomlib.lock")"
+    printf '    "circom-witness-rs": "%s"\n' "$witness"
+    printf '  }'
+    for stem in $STEMS; do
+      printf ',\n'
+      printf '  "%s": {\n' "$stem"
+      printf '    "setup": "%s",\n' "$(stem_setup "$stem")"
```
Fantoni0: This seems to try to read previous setups?

deployments/scripts/circuit-artifacts.sh
```diff
@@ -0,0 +1,232 @@
...
+      printf '    "r1cs": "%s",\n' "$(sha256_file "$(file_for "$stem" r1cs)")"
+      printf '    "graph.bin": "%s",\n' "$(sha256_file "$(file_for "$stem" graph.bin)")"
+      printf '    "proving_key.bin": "%s"\n' "$(sha256_file "$(file_for "$stem" proving_key.bin)")"
+      printf '  }'
+    done
+    printf '\n}\n'
+  } >"$LOCK"
```
Fantoni0, line 171: But I think the redirect to lockfile clears contents before any command is run in the block. So that means that we will never be able to read previous setups.?

Fantoni0 (commented):
Left one question, not sure if I read the script right, but I think we might be deleting before reading.

v0-e:
Moved lockfile reads to a pre-loop :+1:

Fantoni0 (approved):
LGTM!

PR #499, opened 2026-08-21
SDK pull artifacts

v0-e:
Adds the `CircuitStore` helper to support (down)loading the circuit artifacts.
Adds circuits.json to `sdk/native` so that it is inside the crate dir, so it can be loaded up during build and allowing it to be published.


Fantoni0 (approved):
LGTM 🚀

PR #500, opened 2026-08-21
SDK crate `Cargo.toml` publish flag as true

v0-e:
Make the crate publishable!

`cargo publish -n` works in current state, so we can already publish the SDK. But better way for the #499 stack to be included.


v0-e:
Perhaps we can ignore the semver check here, it is related with the crate renaming.

Fantoni0 (approved):
LGTM!

PR #501, opened 2026-08-21
Remove CI `e2e-testnet` environment

v0-e:
Removes that environment, which now is mostly a leftover of e2e web tests as we have ephemeral accounts.
Only used for `E2E_FREIGHTER_PASSWORD`, but we can generate it per run and pass it through `$GITHUB_ENV`, so no stored secret is needed.


cupicmarko (approved):
Sorry, forgot about that when I reactored account provisioning.

PR #506, opened 2026-08-24
SDK GVK crypto

v0-e:
SDK/off-chain crypto for GVK support. BabyJubJub arithmetic and encryption. Mostly ports from `circuits/`.


sdk/native/src/zk/babyjub.rs
```diff
@@ -0,0 +1,172 @@
...
+    let gamma = p.y * q.x;
+    let delta = (p.y - a * p.x) * (q.x + q.y);
+    let tau = beta * gamma;
+
+    let x_den = (Scalar::one() + d * tau)
+        .inverse()
+        .expect("BabyJubJub addition denominator is never zero");
+    let y_den = (Scalar::one() - d * tau)
+        .inverse()
+        .expect("BabyJubJub addition denominator is never zero");
```
Fantoni0: This will only be true if p and q are in the curve. Maybe would make sense to check for that and returning a Result rather than panicking?

sdk/native/src/zk/gvk.rs
```diff
@@ -0,0 +1,382 @@
...
+            amount: Field::from(amount),
+            blinding,
+            salt,
+        }
+    }
+
+    /// Encrypt this note under authority key `D`, nonce, and per-note index.
+    pub fn encrypt(
+        &self,
+        admin_pub_key: &BabyJubJubPoint,
```
Fantoni0, line 67: We are not making any of the subgroup checks for the admin_pub_key (D) that we do inside the circuit. 
Since now it's part of the SDK it would be better to add some checks here. Let me know if I can help

sdk/native/src/zk/gvk.rs
```diff
@@ -0,0 +1,382 @@
...
+        poseidon2_hash3_internal(h3, idx, Scalar::from(0u64), dom)
+    }
+}
+
+impl GvkRecoveredNote {
+    pub fn amount(&self) -> NoteAmount {
+        let bytes = self.amount.to_le_bytes();
+        let mut amount_le = [0u8; 16];
+        amount_le.copy_from_slice(&bytes[..16]);
+        NoteAmount::from(u128::from_le_bytes(amount_le))
```
Fantoni0: We are silently truncating here.

sdk/native/src/zk/babyjub.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+}
+
+/// Point doubling, matching circomlib's `BabyDbl`.
+pub fn double(p: Point) -> Point {
+    add(p, p)
+}
+
+/// Multiply `point` by the integer value of `scalar` via double-and-add.
+pub fn scalar_mul(point: Point, scalar: Scalar) -> Point {
```
Fantoni0: Let's add a comment signaling this is not constant time. Fine for our use cases, but let's document it

sdk/native/src/zk/gvk.rs
```diff
@@ -0,0 +1,382 @@
...
+        let inputs = if gvk_mode == GvkMode::Traceable {
+            Some(
+                input_notes
+                    .iter()
+                    .enumerate()
+                    .map(|(k, note)| note.encrypt(&admin_pub_key, &nonce, k))
+                    .collect::<Result<Vec<_>>>()?,
+            )
+        } else {
+            None
```
Fantoni0, line 192: Maybe adding a print here if inputs >0 for non-traceable mode. To signal the SDK user he shouldn't be passing inputs for that config

Fantoni0 (changes requested):
Thanks @v0-e 💯 
I left some comments and nitpicks. I know many of these were already present in the circuit crate. But there it was mainly used for hardcoded test vectors, and now this will be part of the SDK so I wanted to add some defensive checks.
Feel free to ignore some if you don't agree

sdk/native/src/zk/gvk.rs
v0-e: Maybe error is best, added it :+1:

sdk/native/src/zk/babyjub.rs
v0-e: Added it :+1:

sdk/native/src/zk/gvk.rs
v0-e: Should be ok, but made it fallible :+1:

sdk/native/src/zk/gvk.rs
v0-e: Indeed, that is missing, added :+1:

sdk/native/src/zk/babyjub.rs
v0-e: Indeed, made it fallible (`Option`s) :+1:

v0-e:
Thank you @Fantoni0, I agree with all comments 🙌 
Let me know if any check is missing!

Fantoni0 (approved):
LGTM! 👍

PR #507, opened 2026-08-24
fix(pool): report verifier rejections as InvalidProof, and pin UnknownRoot + zero-value

zexoverz:
Towards #374 (stale ASP roots, zero-value edge cases, malformed public inputs). Four negative tests on the transact path: a root never inserted, a root evicted from the history ring, `UnknownRoot` reported ahead of the nullifier and ext-hash checks, and zero-value behaviour. Writing them surfaced a real defect, so the fix is here too. `Groth16Error` and the pool `Error` are separate `#[repr(u32)]` enums with overlapping codes — `MalformedPublicInputs` is 1 and so is `NotAuthorized`, `MalformedProof` is 2 and so is `MerkleTreeFull`. `verify_proof` called the verifier without `try_`, so a rejection trapped out of the pool frame carrying the verifier own code and the caller decoded it against the pool enum: a refused proof arrived as an authorization failure. The verifier also never answers `Ok(false)` — every rejection path returns `Err(Groth16Error)`, including the pairing check — so `Error::InvalidProof` was unreachable from proof verification and the `if !Self::verify_proof(..)` branch was dead code. The fix is one expression, `try_verify` with any verifier error mapped to the existing `InvalidProof`, no new enum variant.

Every test asserts a specific error rather than `is_err()`, because the pre-existing `transact_rejects_unknown_root` shows why that matters: its fixture also carries a mismatched `ext_data_hash`, so deleting the root check entirely leaves it green via `WrongExtHash`. Reverting only the `pool.rs` change turns three tests red with `left: Ok(NotAuthorized)` / `right: Ok(InvalidProof)`, and the new `transact_reports_verifier_rejection_as_invalid_proof` has a fixture that excludes every other source of `InvalidProof` on that path. The existing e2e test `transact_rejects_tampered_output_commitment` already asserted this leak against real proofs and now passes. `cargo test --workspace` is 445 passed / 0 failed, fmt and clippy `-Dwarnings` clean, and the changed tests pass under Miri. One follow-up left out deliberately: `contracts/pool-gvk/src/pool_gvk.rs:593` has the identical bug, but the same one-line fix there collides with a negative assertion in `transact_skips_asp_root_canonical_validation_when_flags_ignore_field` and needs a design call about a distinct ASP-root error.

### DIFF
--- a/contracts/pool/src/test.rs
+++ b/contracts/pool/src/test.rs
@@ -1119,14 +1119,10 @@ fn transact_accepts_deposit_at_maximum_bound() {
 
 /// An all-zero proof must be refused with a clean error rather than panicking.
 ///
-/// It is refused, but note what the caller actually receives. `verify_proof`
-/// calls the verifier contract, and `Groth16Error::MalformedPublicInputs` is 1,
-/// which is the same numeric code as this contract's `Error::NotAuthorized`.
-/// The cross-contract error therefore surfaces at the pool boundary as an
-/// authorization failure. Nothing is unsound about it, but the error a caller
-/// sees does not describe what happened, and no existing test reached the
-/// verifier so nothing pinned it. This test pins it, and will fail if the codes
-/// are ever separated, which is the point.
+/// The points are all-zero but not empty, so the `is_empty` guard does not
+/// catch them and the proof reaches the verifier, which refuses it. The caller
+/// sees `InvalidProof`, the pool's own error, rather than whichever
+/// `Groth16Error` the verifier happened to raise.
 #[test]
 fn transact_rejects_zeroed_proof() {
     let env = test_env();
@@ -1154,8 +1150,8 @@ fn transact_rejects_zeroed_proof() {
         .expect_err("a zeroed proof must be refused");
     assert_eq!(
         err,
-        Ok(Error::NotAuthorized),
-        "current behaviour: the verifier's error code 1 is read as the pool's error code 1"
+        Ok(Error::InvalidProof),
+        "a proof the verifier refuses must be reported as InvalidProof"
     );
 }
 
@@ -1331,11 +1327,9 @@ fn transact_reports_unknown_root_before_later_checks() {
 /// so the amount checks pass too and the transaction is refused only by the
 /// verifier.
 ///
-/// The error that surfaces is `NotAuthorized` rather than `InvalidProof`:
-/// `Groth16Error::MalformedPublicInputs` is 1, the same numeric code as this
-/// contract's `NotAuthorized`, so the verifier's error is remapped at the
-/// contract boundary. Nothing is unsound about it, but it is pinned here so
-/// that separating the codes is a deliberate change.
+/// The error that surfaces is therefore `InvalidProof` and nothing earlier,
+/// which is what makes this a statement about the deposit bound rather than
+/// about the proof.
 #[test]
 #[cfg_attr(miri, ignore)]
 fn transact_accepts_zero_ext_amount_with_zero_maximum_deposit() {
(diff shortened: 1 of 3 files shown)


zexoverz:
Rebased past the conflict, which was #485 landing in the same file — both sides were additive so both blocks are kept. 32/32 pool tests green locally; the workflows here are sitting at `action_required` and need an approval to run.

zexoverz:
Expanded scope: the negative tests turned up a real error-mapping bug, so the one-expression fix is in here rather than a follow-up PR. Details in the description; `pool-gvk` has the same bug but needs a design call, so it is left out.

zexoverz:
Also fixes the nightly `pool` UB group, which went red after #485. Two of those tests abort under Miri: one is cured by the `try_verify` change here since the rejection no longer traps out of the frame, the other gets the ignore attribute its neighbours already carry. `cargo miri test --lib -p pool` is 16 passed / 0 failed.

contracts/pool/src/test.rs
```diff
@@ -1168,7 +1164,15 @@ fn transact_rejects_zeroed_proof() {
 /// by the circuit (`transaction.circom` constrains it), and the contract relies
 /// on that rather than re-checking independently. The test is written to fail
 /// if that reliance ever changes, in either direction.
+///
+/// Ignored under Miri, matching every other `transact_*` test in this file. The
+/// abort is Stacked-Borrows UB inside the host's own error path, not in this
+/// contract, and it is what turned the nightly `pool` group red after #485.
```
Fantoni0: In general, let's try to keep AI documentation verbosity on the low end. Let's not mention PR or issue numbers either

contracts/pool/src/test.rs
```diff
@@ -1197,3 +1201,217 @@ fn transact_leaves_duplicate_nullifier_detection_to_the_circuit() {
...
+///
+/// - the two ASP-root comparisons are skipped, because the pool is registered
+///   with policy flags 0 and both are behind `requires_*_proofs`;
+/// - the `is_empty` guard in `verify_proof` does not fire, because the mock
+///   proof carries non-zero points, asserted below.
+///
+/// Everything before proof verification is made to pass: the root is the pool's
+/// live root, the nullifier is unspent, the ext hash is computed from the ext
+/// data, and the public amount is zero for a zero-value transaction. So the
+/// verifier is the only thing left that can refuse this transaction.
```
Fantoni0: Same as above, let's avoid 3 paragraph comments for a test

contracts/pool/src/test.rs
```diff
@@ -1197,3 +1201,217 @@ fn transact_leaves_duplicate_nullifier_detection_to_the_circuit() {
...
+    let err = pool
+        .try_transact(&proof, &ext, &Address::generate(&env))
+        .expect_err("a proof the verifier refuses must be refused by the pool");
+    assert_eq!(
+        err,
+        Ok(Error::InvalidProof),
+        "a verifier rejection must be reported as the pool's InvalidProof, not as the \
+         NotAuthorized that Groth16Error::MalformedPublicInputs shares a code with"
+    );
+}
```
Fantoni0, line 1376: I would like to see a test where ext_amount > 0 (deposit) and we have an invalid proof, so we can verify that everything reverts correctly and no funds are taken from the user.

Fantoni0 (changes requested):
Looks good, thanks @zexoverz ! 🚀 
Just left some comments related to stylism, and maybe adding some additional test to  check all possible paths.

Also, thanks so much for your contributions torwards #374 🙌 . But in the future, if you want to contribute please take a look to issues with "help-wanted", "contributors-friendly" or "good-first issue" tags 🙏

zexoverz:
Comments trimmed and the issue reference dropped. Added `transact_rejects_deposit_with_invalid_proof_without_moving_funds`: a 500-unit deposit with a proof the verifier refuses, asserting `InvalidProof` and that neither balance moves. It registers a real asset contract rather than `MockToken`, whose `balance` always returns 0 and whose `transfer` is a no-op, so the assertion is an actual observation. 34/34 green, and noted on the tags.

Fantoni0 (approved):
LGTM!

PR #510, opened 2026-08-25
Split the note-owner address from the signing address

cupicmarko:
### What this is

One `String` in the SDK was doing three unrelated jobs at once:

- selecting the notes and key material of the account that **owns** them
- supplying the pool contract's `sender` argument
- being the transaction envelope's **source** — the account that signs and pays

The first is one identity. The second and third are another. Nothing in the type system distinguished them, so nothing could catch one being used where the other belonged.

This adds `NoteOwnerAddress` and `SignerAddress` as distinct types and threads a second address from the account-opening boundary (`Client::account`) through `Account`, `PrivatePoolConfig`, both `Signer` implementations, the wasm bridge and the JS wrapper.

**Every caller sets the signing address to the note owner, so nothing behaves differently.** This is plumbing only.

### Why

The pool's privacy depends on the account owning a note not being the account that appears on the ledger spending it. Every Stellar transaction publicly names its source account; if that source is also the note owner, an observer links the two and the pool has bought the user nothing.

Separating those identities is the feature. This PR does the part that must come first — making them separate *values* — so a later change can give them different values without rewriting every call site.

### Why types, not a second `String`

They're load-bearing. The first build after introducing them rejected 18 call sites, three genuinely misclassified. With two bare strings those would have compiled and failed silently at runtime — signing as the note owner while the caller believed otherwise, which is precisely the link this feature exists to break. A wrong address here doesn't error; it produces a valid transaction
attributed to the wrong identity.

There is deliberately **no `From` conversion between the two types**. Code that wants both to be the same account constructs each from the same string, where review can see it. `Deref` is withheld for the same reason.

`prepare_pool_transact` takes `&SignerAddress` rather than `&str`, so the one call site where the note-owner identity reached the on-chain roles can no longer be handed the wrong address. That parameter feeds the contract sender, the sequence-number lookup and the envelope source at once.

### Two things that look like oversights and aren't

- **`prepare_register` keeps the note-owner address.** Which identity should   source a public-key registration is an open question no requirement settles —  the value there is both the registry key and the transaction source.
- **`require_funded_for_pool` also keeps it**, for a different reason: a deposit pulls tokens from whoever sends the transaction, so it's the one operation where the two identities *cannot* diverge.

### Scope

No UI, no way to choose a different signer, no `ExtData` change. The ASP admin page is untouched — its wallet is a third identity, neither note owner nor signer, and no requirement covers it.

### Reviewing

The diff is large but mechanical, and the compiler verified the classification rather than review having to. The parts worth attention: `types/address.rs` (why the conversions are withheld), the single changed call site in `pool.rs`, and `sdk/web/src/signer.rs`.


Fantoni0 (dismissed):
I was going to leave multiple comments 👀 Until I saw the description :P 
LGTM! Thanks @cupicmarko 🙌

Fantoni0 (dismissed):

Fantoni0 (approved):

PR #511, opened 2026-08-25
SDK GVK wiring

v0-e:
Adds all related GVK-related logic to the SDK.
The artifacts and web support are provided in following PRs.


sdk/native/src/gvk/audit.rs
```diff
@@ -0,0 +1,500 @@
...
+        let mut nullifiers = Vec::with_capacity(group.nullifiers.len());
+        let mut audited_inputs = Vec::new();
+        for (nullifier, ciphertext) in group.nullifiers {
+            nullifiers.push(nullifier);
+            if let Some(ct) = ciphertext {
+                let d_priv = self.d_priv;
+                let candidates = self.commitment_candidates().await?;
+                if let Some(note) = try_decrypt_against_commitment_set(&d_priv, &ct, candidates) {
+                    audited_inputs.push(note);
+                }
```
Fantoni0: I am afraid we might loose the relation between nullifiers and inputs as they are in separate lists. If this fails for whatever reason (not sure if we keep dummy 0 values notes), the 1-to-1 relation between nullifiers and inputs breaks I think

sdk/native/src/pool.rs
```diff
@@ -263,6 +264,29 @@ impl<S: Storage> PrivatePool<S> {
         Ok(())
     }
 
+    pub async fn audit(&self, global_view_private_key: Field) -> Result<GvkAudit<S>, Error> {
+        let pool = self
+            .config
+            .contract_config
+            .pool(&self.config.pool_contract_id)
+            .map_err(|e| Error::InvalidConfig(e.to_string()))?;
```
Fantoni0, line 272: should we also check that `global_view_private_key` and the pool key match?

sdk/native/src/state/storage.rs
```diff
@@ -929,14 +948,105 @@ impl Storage {
...
+            match kind {
+                0 => {
+                    let encoded = encoded.ok_or_else(|| {
+                        anyhow!("commitment row missing gvk_ciphertext after filter")
+                    })?;
+                    Ok(crate::gvk::GvkEvent::Commitment {
+                        ledger,
+                        event_id,
+                        commitment: field,
+                        gvk_ciphertext: decode_gvk_ciphertext(&encoded)?,
```
Fantoni0: I think we are double-decoding here? I think line 1010 already decodes

sdk/native/src/zk/flows.rs
```diff
@@ -818,6 +901,40 @@ where
...
     circuit.set_single("extDataHash", &be32_to_0x_hex(&ext_data_hash_be));
+    if gvk_mode != GvkMode::Off {
+        circuit.set_single("nonce", &be32_to_0x_hex(&ext_data_hash_be));
+    }
+
+    let (output_gvk_ciphertexts, input_gvk_ciphertexts) = if gvk_mode == GvkMode::Off {
+        (None, None)
+    } else {
+        let admin = admin_view_key.as_ref().expect("validated above");
+        let nonce = Field(U256::from_big_endian(&ext_data_hash_be));
```
Fantoni0: Let's use Field::try_from as it ensures the U256 is in range

sdk/native/src/gvk/audit.rs
```diff
@@ -0,0 +1,500 @@
...
+pub struct GvkTxAudit {
+    pub ledger: u32,
+    pub outputs: Vec<GvkAuditedNote>,
+    /// Traceable pools only; empty for view-only pools.
+    pub inputs: Vec<GvkAuditedNote>,
+    /// Public nullifiers emitted in this transaction.
+    pub nullifiers: Vec<Field>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
```
Fantoni0: Not sure if Debug can mess with the telemetry and Sensitive tags that Marko introduced .Ideally this should be secret and not printed in regular prints

Fantoni0 (changes requested):
Great @v0-e 
Left some comments

sdk/native/src/gvk/audit.rs
v0-e: Indeed :+1:

sdk/native/src/pool.rs
v0-e: Added :+1:

sdk/native/src/state/storage.rs
v0-e: Yeah :+1:

sdk/native/src/zk/flows.rs
v0-e: Fixed it :+1:

sdk/native/src/gvk/audit.rs
v0-e: I remove `Debug` for now, but perhaps we should review how best to do this across the repo. There are other types that have raw `Debug` derives.

Fantoni0 (approved):
LGTM!

PR #513, opened 2026-08-25
GVK artifacts

v0-e:
Adds GVK circuit artifacts.
- Renames circuit files to `policy_tx_2_2_[A|&B]_gvk_[V|T]`;
- Bump lockfile `circuits.json` to `0.3`.


sdk/native/src/types/circuit_stem.rs
```diff
@@ -70,53 +69,31 @@ impl fmt::Display for CircuitStem {
...
-    let policy_flags = if suffix.is_empty() {
-        PolicyFlags::EMPTY
-    } else {
-        PolicyFlags::from_stem(&format!("{POLICY_TX_2_2}_{suffix}"))?
-    };
-
-    Ok((policy_flags, gvk_mode))
+fn parse_gvk_suffix(stem: &str) -> Option<(&str, GvkMode)> {
+    let (base, mode_word) = stem.rsplit_once(GVK_STEM_SUFFIX)?;
+    let gvk_mode = gvk_mode_from_stem_word(mode_word).ok()?;
```
Fantoni0: Minor, but before we propagated the error, now we simply return None if it fails

cli/src/artifacts.rs
```diff
@@ -23,7 +23,6 @@ pub fn load_transact_artifacts(
 ) -> Result<Vec<(CircuitStem, ProverArtifacts)>> {
     CircuitStem::all_transact_stems()
         .into_iter()
-        .filter(|stem| stem.gvk_mode == GvkMode::Off)
```
Fantoni0, line 26: We no longer filter if Gvk mode is off.  I think this means we load all the GVk circuits even if not needed? Or this was intentional?

sdk/native/src/circuits.rs
Fantoni0, line 113: Not modified, but curious if this now takes more time. Are we excluding GVK circuits for a non GVK pool, or download everything right now?
Maybe can be addressed later, curious to know if you notice a significant increase because of the additional circuits

v0-e: It will take a bit more time to download. We only have one compressed file for all the artifacts, so there is no way to choose which ones to download. I don't think it is important for now.

I do agree that this could be improved. My biggest issue is not the download size, but the amount of circuits that the user is going to have and not use. Perhaps the best approach would be to split the release bundle into sub-bundles per stem. Each sub-bundle would have the proving key, r1cs, witness graph for that circuit. We can publish several files per circuits release in Github Releases :+1:

sdk/native/src/types/circuit_stem.rs
v0-e: :+1: improved that

cli/src/artifacts.rs
v0-e: Yes, the CLI loads everything regardless. It was intentional. While testing the CLI for other stuff, I found no impact to the UX.
This would be part of another related improvement: CLI selective loading (select enabled pools) or lazy loading. I mentioned that for later work [here](https://github.com/NethermindEth/stellar-private-payments/pull/521#issue-5270045297).

Fantoni0 (approved):

PR #514, opened 2026-08-25
WASM/Web GVK support

v0-e:
Adds GVK support to the web libraries. This mainly exposes GVK admin types since user GVK logic is already handled by the base SDK.

Also improves a bit the circuit hashes handling to both SDKs as we add more stems. #535 touches related logic also.


sdk/web/src/workers/prover.rs
```diff
@@ -223,43 +142,33 @@ async fn ensure_disclosure_prover(n_notes: usize) -> Result<(), JsError> {
...
-    r1cs_bytes: &[u8],
-) -> Result<Vec<u8>, JsError> {
-    get_or_derive_uncompressed(pk_name, pk_sha256, move || async move {
-        let compressed = fetch_circuit_file_verified(pk_name, pk_len, pk_sha256).await?;
-        let tmp = Groth16Prover::new(&compressed, r1cs_bytes).map_err(|e| {
+async fn uncompressed_pk_bytes(stem: &str, r1cs_bytes: &[u8]) -> Result<Vec<u8>, JsError> {
+    let pk_name = artifact_file_name(stem, ArtifactKind::ProvingKey);
+    let pk_name_for_closure = pk_name.clone();
+    let stem = stem.to_string();
+    let r1cs_bytes = r1cs_bytes.to_vec();
```
Fantoni0, line 169: We went from a borrow to a to_vec() call over the large pk. Is this intentional?

sdk/web/src/client/gvk.rs
```diff
@@ -0,0 +1,38 @@
...
+#[wasm_bindgen]
+pub struct GvkAudit {
+    inner: NativeGvkAudit<StorageBridge>,
+}
+
+#[wasm_bindgen]
+impl GvkAudit {
+    /// Fetch and audit the next transaction, or `null` when exhausted.
+    #[wasm_bindgen(js_name = nextTx)]
+    pub async fn next_tx(&mut self) -> Result<JsValue, JsError> {
```
Fantoni0: Can we have a race condition here if an auditor tries to call this twice?

sdk/web/src/workers/storage.rs
```diff
@@ -939,22 +954,46 @@ impl Storage for StorageBridge {
...
+                30_000,
+            )
+            .await
+        {
+            Ok(StorageWorkerResponse::PoolCommitmentHashes(hashes)) => Ok(hashes),
+            Ok(other) => Err(Error::Other(format!(
+                "unexpected storage response listing pool commitment hashes: {other:?}"
+            ))),
+            Err(e) => Err(Error::Other(e.to_string())),
+        }
```
Fantoni0: Could we use pagination here?

Fantoni0 (commented):
Great @v0-e 🙌 
Left some questions

sdk/web/src/workers/prover.rs
v0-e: Reverted :+1:

sdk/web/src/client/gvk.rs
v0-e: Added a guard :+1:

sdk/web/src/workers/storage.rs
v0-e: Improved the related flow, we don't load all the commitments now.

v0-e:
Thank you @Fantoni0 🙏
Best we tackle #535 first so we can move the NPM package out of alpha, without GVK support (to match the current published SDK crate).

sdk/web/src/client/pool.rs
```diff
@@ -190,4 +190,16 @@ impl PrivatePool {
...
         .await
     }
+
+    /// Open an admin audit cursor for this pool.
+    ///
+    /// `globalViewPrivateKeyHex` is the admin authority scalar as a
+    /// `0x`-prefixed field hex string. Requires a pool-gvk deployment with
+    /// GVK enabled.
+    pub async fn audit(&self, global_view_private_key_hex: &str) -> Result<GvkAudit, JsError> {
+        let d_priv = Field::from_str(global_view_private_key_hex)
```
Fantoni0: Minor, but all other instances do a trim(), it's missing here

sdk/web/src/client/gvk.rs
```diff
@@ -0,0 +1,52 @@
...
+        if self.busy.replace(true) {
+            return Err(JsError::new("GvkAudit.nextTx() already in progress"));
+        }
+        let _guard = BusyGuard(&self.busy);
+        match self.inner.next_tx().await {
```
Fantoni0: The guard works , but because it'sa #[wasm_bindgen] export, not sure if the first conversion (done by wasm-_-bindgen) will  get the mut reference a lock it. Could you double check 🙏 ?

Fantoni0 (changes requested):
I think it's almost ready @v0-e 
I just have a minor thing and a question related to the wasm export.

sdk/web/src/client/pool.rs
v0-e: Added :+1:

sdk/web/src/client/gvk.rs
v0-e: Nice, looks like that is the case. Using a slightly simpler approach now :+1:

Fantoni0 (approved):
LGTM! Thanks @v0-e 
I think the failing CI is not related to this PR changes.

PR #516, opened 2026-08-25
Release CI

v0-e:
Release CI for:
- Rust SDK (`sdk/native`);
- NPM package (`sdk/web`);
- CLI (`cli/`).

DevEx should be: just bump version on the respective file (e.g., crate `Cargo.toml`, `package.json`). The CI will then run tag current commit on main and publish automatically.

The versioning proposed approach here is to share major/minor version among all release types (at least before `1.x.x`). Why? Simply ensure that **all** released code is interoperable for a given major/minor.
Patch versions are OK to be bumped independently, major/minor versions require a bump across all release types so that they are in sync. 

Joins existing NPM/CLI release CI into the same CI file. A bit of a tradeoff here but useful since the idea is for everything to sync the `X.Y.z` major/minor versions.


.github/workflows/release.yml
```diff
@@ -1,24 +1,170 @@
...
 on:
+  pull_request:
+    branches: ["main"]
+    paths: &release_paths
+      - "sdk/native/Cargo.toml"
+      - "sdk/web/package.json"
+      - "cli/Cargo.toml"
+      - "Cargo.toml"
+      - "scripts/release-versions.sh"
+      - ".github/workflows/release.yml"
```
Fantoni0, line 18: AI complains about this. Syas Github Ci ddoes not use a standard YAML loader and it will make the whole job fail. Can you double check?

Cargo.toml
```diff
@@ -77,7 +77,7 @@ soroban-sdk = { version = "27", default-features = false, features = ["hazmat"]
 soroban-utils = { path = "contracts/soroban-utils" }
 sqlite-wasm-rs = { version = "0.5.5", default-features = false }
 sqlite-wasm-vfs = { version = "0.2.0", default-features = false }
-stellar-private-payments = { path = "sdk/native" }
+stellar-private-payments = { path = "sdk/native", version = "0.1.0" }
```
Fantoni0, line 80: If it's a local dep, do we need to have the version? What happens if someone bumps the SDK version? Also happens in bootnode and cli.

.github/workflows/semver-checks.yml
```diff
@@ -49,24 +47,22 @@ jobs:
       - name: Get latest version tag
         id: get-baseline
         run: |
-          # Find the latest version tag (v0.1.0 or higher)
-          latest_tag=$(git describe --tags --abbrev=0 --match='v*' 2>/dev/null || echo "")
+          # Find the latest sdk release tag (sdk-v0.1.0 or higher)
+          latest_tag=$(git describe --tags --abbrev=0 --match='sdk-v*' 2>/dev/null || echo "")
```
Fantoni0, line 51: Previus releases were vx.y.z. Would this always fail? Or are we creating the first sdk-v version manually?

Fantoni0 (commented):
Looks great 🙌 .
Left some questions

.github/workflows/release.yml
v0-e: CI already ran in this PR. If the syntax was invalid, it probably would have failed the whole job, not sure 🤔 
https://github.com/NethermindEth/stellar-private-payments/actions/runs/33081003746/job/98547882705

Cargo.toml
v0-e: The idea is if we ever publish another crate that depends on the Rust SDK crate, then that crate needs a version also. In this case it would use `0.1.0`.
I analyzed also adding checks for this specific cases, but parsing the file is a bit finicky to be doing it in scripts. Also we don't need that for now.

.github/workflows/semver-checks.yml
v0-e: Yeah, I pushed sdk-v0.1.0 manually on the first crate release commit.
https://github.com/NethermindEth/stellar-private-payments/tree/sdk-v0.1.0

v0-e:
Thank you @Fantoni0 🙏 
I just added an unrelated commit to bump action versions. Github was warning on it using Node v20 vs v24.

.github/workflows/semver-checks.yml
Fantoni0: @v0-e Is this a thing?

v0-e: The manual tag? I did it because the crate had to be published manually for the first time. The CI should do the tags from now on 👍

Fantoni0 (approved):

PR #517, opened 2026-08-26
Fix/increase tree

Fantoni0:
Increases the Merkle tree heights from 10 to 20. This means we go from supporting 1024 leafs (512 transact calls) to 1M leafs (0.5M transact calls):

- Separates levels from ASP levels. Now we can have different levels for the leaves Merkle tree (10) and the ASP membership trees (10).
- Because of that increase, I created the `PrefixTree`. Before, we were building Merkle trees for testing by iteratively calling insert_two_leaves. Which worked for h=10, but it was extremely inefficient for h=20.
- Updates heights, SDK, circuits, and artifacts.
- Redeploys contracts.
- Bumped the circuits artifact version.

Now the demo feels heavier (especially disclosure, which might involve multiple Merkle path proofs): around 12s for a transact call (in my average :smiling_face_with_tear:  single-CPU performance machine). But I think it's an ok tradeoff for supporting real deployment sizes. Let me know your findings and if you think we should lower it a bit (e.g, h=18 would give us 256k leaves and 128k transact calls).
With these new configurations, we shouldn't fill the Merkle tree every couple of PRs.


v0-e (dismissed):
nice!

cupicmarko (approved):

PR #520, opened 2026-08-27
GVK deployment support

v0-e:
Mainly changes to `deploy.sh`.
Also changes second deployed pool to be a GVK traceable, and adds an example.

Added also a key gen tool.


deployments/testnet/deployments.json
```diff
@@ -23,18 +23,20 @@
       }
     },
     {
-      "poolContractId": "CBMRWHTPWJ73LAAPMZQ3Z3CIPD2N5NIBVTPQBG7I7ZEMMHWUHQLFNUVS",
-      "tokenContractId": "CCUUDM434BMZMYWYDITHFXHDMIVTGGD6T2I5UKNX5BSLXLW7HVR4MCGZ",
-      "deploymentLedger": 4343711,
+      "poolContractId": "CCJDARG7BAR34WNXCNHYBMFZJUT6B4XVWARXEIRKAA2IA7UE5X2OCXQB",
+      "tokenContractId": "CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC",
```
Fantoni0, line 27: No EURC pool anymore. Is it intentional?

scripts/e2e-preflight.sh
```diff
@@ -460,14 +460,14 @@ check_env_pool_matches_deployments() {
...
+if not native:
     sys.exit(1)
 print(native[0]["poolContractId"])
 PYEOF
 )"
-  if [ -z "$pool_json" ]; then _STATUS="MISSING"; _DETAIL="could not resolve a single enabled native pool from $deployments_json"; return; fi
+  if [ -z "$pool_json" ]; then _STATUS="MISSING"; _DETAIL="could not resolve an enabled native pool from $deployments_json"; return; fi
   if [ "$pool_env" = "$pool_json" ]; then _STATUS="OK"; _DETAIL="matches deployments.json ($pool_json)"
-  else _STATUS="MISSING"; _DETAIL="E2E_POOL_CONTRACT ($pool_env) does not match the enabled native pool in deployments.json ($pool_json) — a redeploy invalidated the env file"; fi; }
+  else _STATUS="MISSING"; _DETAIL="E2E_POOL_CONTRACT ($pool_env) does not match the first enabled native pool in deployments.json ($pool_json) — a redeploy invalidated the env file"; fi; }
```
Fantoni0: Now we can have more than one native pool. It's not failing, but I think it only works if native[0] == E2E_POOL_CONTRACT. Maybe we could relax the check to a membership check E2E_POOL_CONTRACT \in native?

sdk/native/src/types/gvk.rs
```diff
@@ -179,6 +179,46 @@ impl GvkMode {
     }
 }
 
+/// Operator-stored GVK admin authority key material (local wallet setting)
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GvkAuthoritySetting {
```
Fantoni0: Probably we don't need Debug here as it's private. Probably either drop it or limit it only to he public_key component

tools/gvkey-gen/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+        use std::os::unix::fs::PermissionsExt;
+        file.set_permissions(std::fs::Permissions::from_mode(0o600))?;
+    }
+    file.write_all(contents.as_bytes())?;
+    Ok(())
+}
+
+fn open_storage(path: &std::path::Path) -> Result<LocalStorage> {
+    LocalStorage::open(&path.to_string_lossy())
+        .with_context(|| format!("open wallet database at {}", path.display()))
```
Fantoni0: If this is called before the --db exists, can it write the file without the proper permissions (the 0600 mode that write-private_file enforces)?

deployments/scripts/deploy.sh
```diff
@@ -373,24 +511,24 @@ get_latest_ledger_seq() {
...
-build_verifier_wasm_for_suffix() {
-  local suffix="$1"
+build_verifier_wasm_for_key() {
+  local key="$1"
   local wasm_name vk_path tmp_vk=""
-  wasm_name="$(verifier_wasm_name_for_suffix "$suffix")"
+  wasm_name="$(verifier_wasm_name_for_key "$key")"
 
-  if [[ "$suffix" == "AB" && -n "$VK_FILE" ]]; then
+  if [[ "$key" == "AB" && -n "$VK_FILE" ]]; then
```
Fantoni0: I think this needs to be updated for AB_gvk_ pools?

deployments/scripts/deploy.sh
```diff
@@ -373,24 +511,24 @@ get_latest_ledger_seq() {
...
     [[ -f "$VK_FILE" ]] || die "vk file not found: $VK_FILE"
     vk_path="$VK_FILE"
-  elif [[ "$suffix" == "AB" && -n "$VK_JSON" ]]; then
+  elif [[ "$key" == "AB" && -n "$VK_JSON" ]]; then
```
Fantoni0: Same as above

Fantoni0 (changes requested):
Great @v0-e . Left some questions

deployments/scripts/deploy.sh
v0-e: Fixed :+1:

scripts/e2e-preflight.sh
v0-e: Added that check :+1:

sdk/native/src/types/gvk.rs
v0-e: Removed it :+1:

tools/gvkey-gen/src/main.rs
v0-e: Added that :+1:

deployments/testnet/deployments.json
v0-e: Yes, not sure if we can support >2 pools rn (see #544 for a fix). But anyway EURC / non-XLM tokens, deserve a review due to the units resolution (iirc 10^6 instead of 10^7), that is why it was allowlist in the policy. 
For current demos, XLM blocklist and XLM blocklist + traceable are best for now. If I had to pick two configs.

Fantoni0 (approved):
LGTM! @v0-e  💯

PR #521, opened 2026-08-27
Web lazy load transact artifacts

v0-e:
Enables lazy loading for transact artifacts in the prover web worker. Acts now the same as loading disclosure artifacts.
Also simplifies init of that worker, because now it is ready instantly.

Lazy loading in the CLI would also be interesting, but everything there is generally faster. Leaving that for another PR.


sdk/web/src/client/e2e_tests.rs
```diff
@@ -297,12 +297,6 @@ async fn e2e_smoke_client_construction() {
 
     let mut client = build_test_client(&storage).await;
 
-    // `Client::new` spawns the prover but does not ping it; ping explicitly.
-    client
-        .ensure_prover()
-        .await
-        .expect("prover worker must start and answer its ping");
-
```
Fantoni0, line 305: Does removing this affect the thing the smoke test is checking?

v0-e: It does test less now. We can keep it as a basic liveness check as in: fast check if the WASM SDK + workers in a browser is loadable.
But we don't lose much by remove it, lmk. Other tests are lengthier but do this anyway.

Fantoni0: Let's keep it

Fantoni0 (approved):

PR #522, opened 2026-08-28
Fix/consolidate wallet interactions

cupicmarko:
## Summary

Hardens the boundary between the app and the Freighter/SEP-0043 wallet across four guarantees, then fixes the defects an independent review found in that work.

**Identity** — splits the previously-conflated "user address" into `NoteOwnerAddress` and `SignerAddress`, removing the implicit conversions between them; the native SDK now refuses any session where the two diverge.

**Signature attribution** — the wallet's self-reported signer is compared against the account actually asked to sign, at the Rust wasm layer and both JS call sites, with a mismatch raised as a distinct, non-rejection error.

**Storage-worker binding** — the browser storage worker is bound to one wallet session for its lifetime; requests touching private key material are refused for any other account, and the raw RPC surface refuses them outright.

**Session lifecycle** — the app survives a mid-session wallet account or network switch: an account change re-binds onboarding to the switched-to account, a network change or unreadable wallet ends the session and returns to a connect prompt.

Supporting changes: a migration adding a uniqueness constraint on derived keypairs per account, per-account scoping of settings, redaction of addresses/amounts in error paths that reach the UI, removal of a note commitment from the URL fragment, and the admin page brought onto the same hardened wallet path it previously bypassed.

## Remediation

An independent review of the above found two blockers and a set of majors, both now fixed:

- Fresh-account onboarding crashed at the disclaimer step (`storage is not defined`); the storage-worker session release called a method the published wrapper didn't expose, so a binding could never be revoked.
- The signature-attribution guard was conditional and silent at all three layers; the storage-worker's session binding was unauthenticated and its privilege list hand-maintained; several session-lifecycle races (publication before supersession check, drain excluding the history write, duplicate unreadable-poll counting, stale runtime after a rebind) were closed; onboarding steps now hold an explicit ownership token instead of inferring abandonment from an error code.

## Testing

- `cargo test --workspace`, `cargo clippy --all-targets -- -D warnings`: clean.
- `npm run test:unit --prefix e2e-freighter`: 202/202 passing.
- `make freighter-e2e`: 14/14 browser specs passing, plus 3 new specs covering fresh onboarding, mid-wizard rebind, and cancellation.


sdk/native/src/client.rs
```diff
@@ -143,13 +143,11 @@ impl<S: Storage> Client<S> {
     )]
     pub fn account(
         &self,
-        user_address: impl Into<NoteOwnerAddress>,
+        user_address: NoteOwnerAddress,
+        signer_address: SignerAddress,
```
v0-e, line 147: Feels like this should collapse to one address. If they must be the same, then let's analyze just passing one address arg. The simpler the API, the best for the user.

v0-e (commented):
Just a small comment so far 😄
The PR is a bit large, if there is anyway to split, it would be great 🙏

PR #533, opened 2026-08-29
asp-membership: optional inserter role (closes #531)

### LINKED ISSUE #531: Pluggable admission into the Association Set: an inserter role on asp-membership, and a KYC-gated admitter that uses it
## Context

The Developer Preview describes the ASP as key-level so that entry "is easier to KYC-gate". Today the only way to gate entry is off-chain: the `asp-membership` admin decides who gets in and calls `insert_leaf`, learning which person each leaf belongs to. If a deployer wants an on-chain admission policy (a contract that verifies something before inserting), the only way to plug it in is `update_admin(policy_contract)`, which hands the whole admin role (including `set_admin_insert_only` and future admin powers) to the policy.

## Proposal (small, in this repo)

Add an optional **inserter** to `asp-membership`:

- `set_inserter(inserter: Option<Address>)`, admin only; `get_inserter()`.
- In `insert_leaf`, when `admin_insert_only` is true: require the inserter's auth if one is set, otherwise the admin's, as today. `None` restores the current behaviour. Open insertion (`admin_insert_only = false`) is unchanged.

That is the whole change: one storage key, two functions, a two-line branch in `insert_leaf`. The admin keeps the admin role; an admission policy contract gets exactly one capability, inserting. Draft PR follows this issue.

## What it enables (reference implementation, outside this repo)

I have a KYC-gated admitter running against your `asp-membership` on testnet (instance built from `cb79f817`): it inserts a leaf only after a hash-based STARK proof that the requester belongs to the issuer's KYC'd set, without revealing which member, with commit-then-reveal so a published proof cannot be replayed with someone else's leaf. Because there was no inserter role, the e2e had to transfer admin to it. Receipts (testnet): handover `28c9ebdc`, admission `47566736` and `844eac22` (LeafAdded, root moved), substitution / replay / tampered proofs rejected, a successor generation refuses old proofs. Code and limits: https://github.com/Galmanus/spp-compliance-layer/tree/main/asp-admitter

With the inserter role, that admitter (or any other policy: allow-list signer, multisig, rate limiter) plugs in with `set_inserter(policy)` and the deployer keeps admin.

## Open question for the maintainers

Whether you would rather express this as a generic pre-insert hook (`asp.insert_leaf` calls `policy.check(leaf)` before inserting) instead of an inserter role. The inserter role is smaller and keeps the proof verification and its gas out of `asp-membership`; the hook keeps the ASP as the single entry point. Happy to do either.

Galmanus:
Closes #531.

## What this adds

An optional inserter role on `asp-membership`, so a deployer can plug an on-chain admission policy into the tree without handing over the admin role via `update_admin`.

- `set_inserter(inserter: Option<Address>)`, admin only. `Some` sets the delegated inserter; `None` clears it.
- `get_inserter() -> Option<Address>`.
- In `insert_leaf`, when `admin_insert_only` is true: the insertion is authorized by the inserter if one is set, otherwise by the admin, as before. Open insertion (`admin_insert_only = false`) is unchanged.

The whole change is one storage key, two functions, and a small branch in `insert_leaf`.

## Backward compatibility

Existing deployments have no `Inserter` key, so `get_inserter()` returns `None` and `insert_leaf` authorizes against the admin exactly as it does today. No migration, no behavioural change unless a deployer opts in with `set_inserter`.

## What it enables

The KYC-gated admitter described in #531 currently has to receive the full admin role to insert (the e2e transfers admin to it). With this, the deployer keeps admin and grants the policy contract exactly one capability, inserting: `set_inserter(policy)`. Any admission policy fits the same slot: an allow-list signer, a multisig, a rate limiter, or the STARK-gated admitter.

## Tests

Seven unit tests on `asp-membership`:
- inserter defaults to `None`
- set and clear (`Some` then `None`)
- `set_inserter` requires admin
- the inserter authorizes an insertion (auth mocked on the inserter, not the admin)
- the admin alone cannot insert once an inserter is set
- clearing the inserter restores admin-authorized insertion
- permissionless insertion ignores the inserter

## Design note

#531 raised two options: an inserter role, or a generic pre-insert hook (`asp.insert_leaf` calls `policy.check(leaf)`). This PR implements the inserter role, which keeps the proof verification and its gas out of `asp-membership`. Happy to rework it as a hook if you'd prefer the ASP to remain the single entry point.

## Verification

`rustfmt` clean. I did not run the full workspace test suite locally (a cold Soroban build exceeded the disk on my machine); the change is a small, test-gated addition plus the storage/auth branch, and CI should cover the suite.

### DIFF
--- a/contracts/asp-membership/src/lib.rs
+++ b/contracts/asp-membership/src/lib.rs
@@ -28,6 +28,10 @@ enum DataKey {
     Root,
     /// Whether admin permission is required to insert a leaf
     AdminInsertOnly,
+    /// Optional delegated inserter authorized to call `insert_leaf` while
+    /// `admin_insert_only` is enabled. When unset, the admin authorizes
+    /// insertions, as before.
+    Inserter,
 }
 
 /// Contract error types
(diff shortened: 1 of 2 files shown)


Fantoni0:
Closed. Context: #531

PR #535, opened 2026-08-31
Configurable deployment, proving keys for web

v0-e:
Allow configuration of the deployment, proving keys for the web SDK.
Also removes `sdk/web/build.rs`, relying on the base SDK / `circuits.json` lockfile for circuit validation.


sdk/web/src/circuits.rs
```diff
@@ -52,7 +56,37 @@ pub(crate) fn ensure_sha256_matches(
...
+}
+
+pub(crate) async fn fetch_circuit_artifact(stem: &str, kind: &str) -> Result<Vec<u8>, JsError> {
+    let lock = stellar_private_payments::circuit_lock().map_err(map_circuit_err)?;
+    let filename = stellar_private_payments::CircuitLockfile::artifact_file_name(stem, kind);
+    let bytes = fetch_circuit_file(&filename).await?;
+    if let Err(err) = lock.verify_artifact(stem, kind, &bytes) {
+        tracing::warn!(
+            "[circuits] hash mismatch for {stem}/{kind}: {err:?}, evicting and refetching"
+        );
```
Fantoni0, line 70: This works at proving time. But since we no longer use build.rs, no checks at compile time exist. Check-wasm-types.js only checks existence, this means we can have R1cs files from older local compiles in target/ (make circuits won't run because we already have it locally) and nothing would prevent us from shipping to web where proving will fail.

CI passes because  wasm-build.yml always runs make circuits.

Fantoni0 (changes requested):

sdk/web/src/circuits.rs
v0-e: Added that check during web build :+1:

Fantoni0 (approved):

PR #536, opened 2026-08-31
GVK admin UI

v0-e:
New view in `/admin.html`
- Has tx, note, graph views;
- Has ledger / timestamp, note owner pub key filters;
- Download feature;
- Rust example shows basic usage of audit.


app/js/admin-gvk.js
```diff
@@ -0,0 +1,1623 @@
...
+  state.poolGvkMode = pool?.gvkMode ?? null;
+
+  if (reset) {
+    state.audit = null;
+    state.rows = [];
+    state.txCounter = 0;
+    state.exhausted = false;
+    state.filteredVisibleCount = BATCH_SIZE;
+    renderResults();
+  }
```
Fantoni0, line 1477: I think if I select pool A, click sync & audit, and then select pool B for audit, it isn't recomputed? 
I would probably need to manually click the sync & audit button again. This can be error-prone, I feel. You think you are auditing pool B, but you are still filtering and seeing pool A.

app/js/admin-gvk.js
```diff
@@ -0,0 +1,1623 @@
...
+function classifyTx(tx) {
+  const inputSum = sumAuditedAmounts(normalizedInputs(tx));
+  const outputSum = sumAuditedAmounts(normalizedOutputs(tx));
+
+  if (!tx.inputs?.length) {
+    return outputSum > 0n ? 'deposit' : 'transfer';
+  }
+
+  return inputSum > outputSum ? 'withdraw' : 'transfer';
+}
```
Fantoni0, line 464: This is failing (deposits are labeled as transfers, see image) and not consistent with `gvk-audit.rs` (which I think is the correct way).
Here, we just check for input length, but deposits have dummy input notes.

Image

Fantoni0 (changes requested):
First of all, great work @v0-e 
Looks awesome, just a couple of minor things. 
Gh does not allow me (or I don't know how) to make general comments and upoad images. I'l leave additional comments on the UI after this.

Fantoni0:
image

I prefer having the initial and the last characters of the address, as we do in Overview

Fantoni0:
image
Filter amount between 5-10 keeps showing a 19 XLM note, and notes smaller than 5

Fantoni0:
image
Note filtering by PK still shows notes by multiple owners (partial matches being shown maybe?)

Fantoni0:
It's hard to read/learn the ledger numbers. Can we make it show the number when you hover?
image

Fantoni0:
This is for the following PRs, not for this one:
The graph works at the note leve, but not sure if it would make more sense to work at the transaction/fund level: We see how funds enter the pool and are disseminated across multiple notes.
Still pretty cool :100: 
image

Fantoni0 (approved):
Amazing 🐐

Issue #167, opened 2026-04-17
Explore efficient Merkle proof building


maksimryndin:
for now we re-build MerklePrefixTree on every transaction (see `build_membership_proof` in `app/crates/platforms/web/src/workers/storage.rs`).

For inactive pools it is absolutely fine to avoid state splits.
But for active pools probably it may take time to re-build each time and we need build once (either on the app load or on the first transaction) and append leaves.

But first we need to measure (e.g. a benchmark) the current approach - how many leaves do correspond to how much Merkle tree build time in a single threaded wasm environment?

related https://github.com/NethermindEth/stellar-private-payments/issues/70

zexoverz:
Taking this one if it is still free. Two things from reading `main` before starting.

The path in the issue is gone. `build_membership_proof` now lives in `sdk/native/src/transact.rs`, so this is the native SDK rather than the single-threaded wasm the issue assumes, and the timings will look different. There are also two rebuild sites in the same call rather than one: `transact.rs:236` for the ASP membership tree and `:313` for the pool tree, both `MerklePrefixTree::new(..).into_built()` per transaction.

There is no bench harness in the repo yet, so I will start with criterion over leaf counts on both sites and post the numbers before proposing anything about caching or appending.

zexoverz:
Numbers first, as promised. Criterion over both rebuild sites, native release, Apple M-series, depths as deployed (`--asp-levels 10`, `--pool-levels 20`).

| leaves | asp d10 | pool d20 |
| --- | --- | --- |
| 0 | 183 µs | 364 µs |
| 16 | 285 µs | 474 µs |
| 256 | 2.55 ms | 2.68 ms |
| 1024 | 9.89 ms | 9.92 ms |
| 4096 | 39.0 ms | 38.2 ms |
| 16384 | 163 ms | 154 ms |

Three things fall out of it.

`MerklePrefixTree::new` is flat in leaf count: 182 µs at zero leaves, 194 µs at 1024, 202 µs at 16384. All of it is the empty-subtree chain, about 9 µs per level. The cost the issue is about is entirely in `into_built`.

`into_built` is linear at roughly 100k leaves/sec, so about 10 µs per leaf. Depth stops mattering above a few hundred leaves, where d10 and d20 converge to within 2%. Below that depth is the whole cost, which is why the two columns differ by exactly 2x at zero.

[...]

PR #537, opened 2026-08-31
bench(sdk): measure the prefix-tree rebuild before changing it

zexoverz:
## 🚀 What's this PR do?

Adds a criterion bench over `MerklePrefixTree::new(..).into_built()` for both rebuild sites, so #167's first question has an answer before anyone proposes a cache. No behaviour changes.

### 📎 Related issues

Refs #167

## 🧠 Context

The issue names one rebuild site at a path that no longer exists. There are two, both in `sdk/native/src/transact.rs`: `:236` for the ASP membership tree and `:313` for the pool tree. Depths here are the deployed ones from `deployments/scripts/deploy.sh` (`--asp-levels 10`, `--pool-levels 20`), not the 8 and 10 the contract and e2e tests use.

Numbers and the reading of them are on the issue. Short version: `new` is flat in leaf count and linear in depth, `into_built` carries the cost at roughly 100k leaves/sec, depth stops mattering above a few hundred leaves, and this runs twice per transaction.

The issue asks for wasm and this is native. The `#[wasm_bindgen_test]` that would produce the wasm figure is written but I could not run it: the documented wasm command fails on macOS because `sqlite-wasm-rs` needs a clang with a wasm backend and Apple clang has none. Detail on the issue. Say the word and I will include it as a target-gated test.

## ✅ Required Checklist

- [x] I've read the [contributing guide](../CONTRIBUTING.md).
- [ ] I've mentioned this PR in the project LinkedIn group — **not done yet, doing this next**
- [x] I've tested this locally. `cargo bench -p stellar-private-payments --bench merkle_build` runs clean. On `.githooks/pre-push` I ran `cargo sort --workspace -g --check` and `cargo shear`, both pass; I did not run `cargo deny` or the JS license audit, so please treat those as unverified by me.
- [x] I've updated or created tests if needed — this is the test

### DIFF
--- a/sdk/native/Cargo.toml
+++ b/sdk/native/Cargo.toml
@@ -75,6 +75,9 @@ rusqlite = { workspace = true }
 hex = { workspace = true }
 sha2 = { workspace = true }
 
+[dev-dependencies]
+criterion = { version = "0.5", default-features = false, features = ["cargo_bench_support"] }
+
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 circuit-keys = { workspace = true }
 contract-types = { workspace = true }
@@ -93,3 +96,7 @@ wasm-bindgen-test = { workspace = true }
 
 [lints]
 workspace = true
+
+[[bench]]
+harness = false
+name = "merkle_build"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,6 +86,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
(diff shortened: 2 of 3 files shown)


sdk/native/Cargo.toml
```diff
@@ -75,11 +75,13 @@ rusqlite = { workspace = true }
 hex = { workspace = true }
 sha2 = { workspace = true }
 
+[dev-dependencies]
+criterion = { version = "0.5", default-features = false, features = ["cargo_bench_support"] }
+
```
Fantoni0: Probably can be moved to the `[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]` section

Fantoni0 (changes requested):
Looks pretty good, thanks for the contribution @zexoverz ! 👍 
Just a minor improvement, and you just need to fix the linter issue and we'll get this merged

zexoverz:
Moved it under the non-wasm dev-deps and fixed the lint, thanks.

Fantoni0 (approved):

PR #538, opened 2026-08-31
fix(pool-gvk): report verifier rejections as InvalidProof

AngeYobo:
## 🚀 What’s this PR do?

Aligns pool-gvk verifier error handling with the existing hardened behavior in pool.

pool-gvk was using client.verify(...), allowing Groth16Error values to cross the contract boundary directly. Because the verifier and pool-gvk error enums use overlapping numeric codes, a verifier rejection such as MalformedPublicInputs could surface as an unrelated pool error such as NotAuthorized.

This PR:

- replaces client.verify(...) with client.try_verify(...)
- maps verifier rejections to Error::InvalidProof
- adds a regression test reproducing the previous NotAuthorized / InvalidProof mismatch
- updates the ASP-root canonicalization test so it only asserts the intended pre-verifier invariant

The previous ASP-root assertion passed for the wrong reason: the verifier error-code collision surfaced NotAuthorized, so !matches!(..., InvalidProof) was vacuously true. The updated assertion now checks the intended NonCanonicalPublicInput invariant directly.

Before the fix:

left:  Ok(NotAuthorized)
right: Ok(InvalidProof)

After the fix, the verifier rejection is reported as Error::InvalidProof.

Validation:

cargo test -p pool-gvk --lib

37 passed
0 failed

The change is limited to pool-gvk verifier error translation and regression coverage. 
No circuit, SDK, deployment, or protocol behavior is changed.

---

### 📎 Related issues (optional)
None

---

## 🧠 Context

`pool` already handles verifier rejections through `try_verify(...)` and maps them to its own `Error::InvalidProof`.

`pool-gvk` still used `verify(...)`, so verifier contract error codes could cross the contract boundary and be decoded against the `pool-gvk` error enum. Because the numeric codes overlap, `Groth16Error::MalformedPublicInputs` could surface as `Error::NotAuthorized`.

This change intentionally mirrors the existing `pool` behavior rather than introducing a new error-handling model.

Risk is limited to error translation: rejected proofs still fail closed before and after this change. No circuit, proof format, or successful transaction path is modified.

---

## ✅ Required Checklist

### DIFF
--- a/contracts/pool-gvk/src/pool_gvk.rs
+++ b/contracts/pool-gvk/src/pool_gvk.rs
@@ -593,7 +593,8 @@ impl PoolGvkContract {
         // `try_verify`, not `verify`. `Groth16Error` and this contract's
         // `Error` are separate `#[repr(u32)]` enums whose codes overlap.
         // Catch verifier rejections here so callers always receive pool-gvk's
-        // own error domain rather than decoding a verifier code as a pool error.
+        // own error domain rather than decoding a verifier code as a pool
+        // error.
         match client.try_verify(&proof.proof, &public_inputs) {
             Ok(Ok(is_valid)) => Ok(is_valid),
             _ => Err(Error::InvalidProof),
--- a/contracts/pool-gvk/src/pool_gvk.rs
+++ b/contracts/pool-gvk/src/pool_gvk.rs
@@ -590,9 +590,14 @@ impl PoolGvkContract {
             }
         }
 
-        let is_valid = client.verify(&proof.proof, &public_inputs);
-
-        Ok(is_valid)
+        // `try_verify`, not `verify`. `Groth16Error` and this contract's
+        // `Error` are separate `#[repr(u32)]` enums whose codes overlap.
+        // Catch verifier rejections here so callers always receive pool-gvk's
+        // own error domain rather than decoding a verifier code as a pool error.
+        match client.try_verify(&proof.proof, &public_inputs) {
+            Ok(Ok(is_valid)) => Ok(is_valid),
+            _ => Err(Error::InvalidProof),
+        }
     }
 
     /// Get the GVK mode, for internal use where the getter's `Result`
--- a/contracts/pool-gvk/src/test.rs
+++ b/contracts/pool-gvk/src/test.rs
@@ -582,9 +582,9 @@ fn assert_policy_transact_skips_ignored_asp_root_validation(flags: u32, nullifie
     assert!(
         !matches!(
             pool.try_transact(&proof, &ext, &sender),
-            Err(Ok(Error::InvalidProof))
+            Err(Ok(Error::NonCanonicalPublicInput))
         ),
-        "flags={flags} should not require ASP root validation for the ignored field(s)"
+        "expected ASP root field to be skipped for flags={flags}"
     );
 }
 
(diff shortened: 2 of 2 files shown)


AngeYobo:
Local validation: pool-gvk tests pass (37/37). I also ran .githooks/pre-push; it currently stops at the cargo-sort check. I reproduced the same cargo sort --workspace --check failure on upstream/main at e1aaaf7, and this PR does not modify any Cargo.toml files.

I’ve also requested access to the project LinkedIn group and will mention this PR there as soon as the membership request is approved.

zexoverz (approved):
Ran it, 37 pass. The `try_verify` shape matches what `pool` does after #507, and the new test is properly non-vacuous, the `!proof.proof.is_empty()` guard being the right detail.

One thing worth a line in the description rather than a silent diff. I put the old assertion back on main and probed what that call actually returns: `Err(Ok(NotAuthorized))`, with `mock_all_auths` on, so it could never have been a real authorization failure. The old `!matches!(.., InvalidProof)` was passing because of the collision rather than despite it. Your change is the first version of that assertion that tests anything.

AngeYobo:
Thanks! agreed. I’ll note that the old assertion passed because of the
enum-code collision, not the intended invariant.


Le mar. 1 sept. 2026 à 00:31, Faisal Firdani ***@***.***> a
écrit :

contracts/pool-gvk/src/test.rs
```diff
@@ -1732,3 +1732,48 @@ fn transact_rejects_replayed_nullifier() {
...
+        VIEW_ONLY,
+    );
+    let pool = PoolGvkContractClient::new(&env, &pool_id);
+    let (member_root, non_member_root) = asp_roots(&setup);
+
+    // Authorization is mocked, so NotAuthorized cannot be a genuine answer.
+    env.mock_all_auths();
+
+    let (proof, ext) =
+        mk_transact_proof(&env, &pool, member_root, non_member_root, 0xE5, VIEW_ONLY);
```
dkgoutham: Since #550, `mk_transact_proof` takes `token: &Address` as its third argument, so this call no longer compiles after a rebase on main. Passing &setup.token fixes it, the same way pool/src/test.rs does.
```
let (proof, ext) = mk_transact_proof(
     &env,
     &pool,
     &setup.token,
     member_root,
     non_member_root,
     0xE5,
     VIEW_ONLY,
);
```

dkgoutham:
Fix is right, same shape as pool after #507. One blocker: branch is 43 behind main, and #550 added a token param to `mk_transact_proof`, so the new test fails to compile once rebased. CI was green against the old base. Please rebase and address the other below comments.

contracts/pool-gvk/src/pool_gvk.rs
```diff
@@ -590,9 +590,15 @@ impl PoolGvkContract {
...
 
-        let is_valid = client.verify(&proof.proof, &public_inputs);
-
-        Ok(is_valid)
+        // `try_verify`, not `verify`. `Groth16Error` and this contract's
+        // `Error` are separate `#[repr(u32)]` enums whose codes overlap.
+        // Catch verifier rejections here so callers always receive pool-gvk's
+        // own error domain rather than decoding a verifier code as a pool
+        // error.
+        match client.try_verify(&proof.proof, &public_inputs) {
```
dkgoutham, line 663: Rejection now returns `Err` after the deposit transfer instead of trapping. Nothing in `pool-gvk` tests the rollback. Port `transact_rejects_deposit_with_invalid_proof_without_moving_funds` from pool?

AngeYobo: Rebased on main, added the token argument, and ported the rollback regression test. pool-gvk passes 44/44 tests; fmt and clippy are clean.

dkgoutham (approved):

PR #540, opened 2026-09-01
Refactor web exports

v0-e:
Reduces hand-written types and provides proper TypeScript types for SDK return values.
The core idea: stop shipping opaque `JsValue` blobs across the WASM boundary and instead expose real, typed classes that `wasm-bindgen` generates TypeScript declarations for automatically.


sdk/web/src/models/config.rs
```diff
@@ -0,0 +1,263 @@
...
+            .map(PoolConfigEntry::from)
+            .collect()
+    }
+
+    /// Plain JSON object matching `deployments.json` (for round-trip input).
+    #[wasm_bindgen(js_name = toJSON)]
+    pub fn to_json(&self) -> Result<JsValue, JsError> {
+        let json = serde_json::to_value(self.native())
+            .map_err(|e| JsError::new(&format!("failed to serialize contractConfig: {e}")))?;
+        serde_wasm_bindgen::to_value(&json)
```
Fantoni0: serde_wasm_bindgen's default Serializer has `serialize_maps_as_objects = false`. So I believe it will be a conflict with these 2 serialize calls as they internally use serialize_map.

sdk/web/src/models/disclosure.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /// Plain JSON object for `JSON.stringify` / receipt export and
+    /// verification.
+    #[wasm_bindgen(js_name = toJSON)]
+    pub fn to_json(&self) -> Result<JsValue, JsError> {
+        let json = serde_json::to_value(&self.inner)
+            .map_err(|e| JsError::new(&format!("failed to serialize disclosure receipt: {e}")))?;
+        serde_wasm_bindgen::to_value(&json)
+            .map_err(|e| JsError::new(&format!("failed to convert disclosure receipt: {e}")))
```
Fantoni0: Same as above

sdk/web/src/client/pool.rs
```diff
@@ -147,15 +154,20 @@ impl PrivatePool {
...
     /// `config` matches [`DisclosureRequest`] (camelCase; `selectedCommitments`
-    /// array with 1..=4 entries). Returns `null` when the account must register
-    /// at the ASP before disclosing.
-    pub async fn disclose(&self, config: JsValue) -> Result<JsValue, JsError> {
+    /// array with 1..=4 entries). Returns `undefined` when the account must
+    /// register at the ASP before disclosing; check with `== null`.
+    pub async fn disclose(&self, config: JsValue) -> Result<Option<DisclosureReceipt>, JsError> {
         with_correlation_id(new_correlation_id(), async {
-            let req: DisclosureRequest = serde_wasm_bindgen::from_value(config)?;
+            let req = DisclosureRequest::from_value(config)?;
```
Fantoni0: Not sure this serde based approach would work. Could we make it use &DisclosureReceipt as verify_disclosure?

sdk/web/src/models/disclosure.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+}
+
```
Fantoni0, line 186: Do we need a from_json method for other functions that require a DisclosureReceipt?

sdk/web/src/models/convert.rs
```diff
@@ -0,0 +1,38 @@
...
+pub(crate) fn gvk_mode_name(mode: GvkMode) -> &'static str {
+    match mode {
+        GvkMode::Off => "off",
+        GvkMode::ViewOnly => "viewOnly",
+        GvkMode::Traceable => "traceable",
+    }
+}
+
+pub(crate) fn policy_flag_names(flags: PolicyFlags) -> Vec<String> {
+    [PolicyFlag::Allowlist, PolicyFlag::Blocklist]
```
Fantoni0: Is it hardcoded instead of iterating over POLICY_FLAGS_IN_SUFFIX_ORDER to avoid the import?

sdk/web/src/models/chain.rs
```diff
@@ -0,0 +1,384 @@
...
+            .map(PoolInfo::from)
+            .collect()
+    }
+
+    #[wasm_bindgen(getter, js_name = aspMembership)]
+    pub fn asp_membership(&self) -> AspMembership {
+        AspMembership::from(self.inner.asp_membership.clone())
+    }
+
+    #[wasm_bindgen(getter, js_name = aspNonMembership)]
```
Fantoni0, line 369: Not sure if these will match the changes in chain_data becaseu of how serde splits the underscore "_"

Fantoni0 (changes requested):
Thanks @v0-e 🙏 
Left some questions and problems I think could surface.
Not a Serde expert by any means, some of these were flagged by AI, but I was able to double-check them. Take a look and let me know your thoughts

v0-e:
Thank you @Fantoni0 🙏 
Please check if the comments were addressed correctly.

sdk/web/src/client/pool.rs
```diff
@@ -146,16 +153,22 @@ impl PrivatePool {
...
-    /// `config` matches [`DisclosureRequest`] (camelCase; `selectedCommitments`
-    /// array with 1..=4 entries). Returns `null` when the account must register
-    /// at the ASP before disclosing.
-    pub async fn disclose(&self, config: JsValue) -> Result<JsValue, JsError> {
+    /// Returns `undefined` when the account must register at the ASP before
+    /// disclosing; check with `== null`.
+    pub async fn disclose(
+        &self,
+        req: &DisclosureRequest,
+    ) -> Result<Option<DisclosureReceipt>, JsError> {
```
Fantoni0, line 161: Looks better now. But we need to make sure that app/js/disclosure has been updated to reflect the new function signature.

sdk/web/src/models/disclosure.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub fn nullifiers(&self) -> Vec<String> {
+        self.inner.nullifiers.iter().map(field_hex).collect()
+    }
+
+    #[wasm_bindgen(getter)]
+    pub fn amounts(&self) -> Vec<String> {
+        self.inner
+            .amounts
+            .iter()
+            .map(|amount| amount.0.to_string())
```
Fantoni0, line 129: I think all the other amoutns are in hex, and now these seem to be base 10

sdk/web/js/index.js
```diff
@@ -158,6 +162,11 @@ export const Storage = { open: openStorage };
 export const Client = {
   new: newClient,
 };
-export { PrivatePool, bootnodeRequired, deriveAspUserLeaf, verifySelectiveDisclosure };
-export { configureTelemetry, set_log_level, dump_recent_logs, debugLogsEnabled };
+export { PoolExecuteResult, PrivatePool, bootnodeRequired, deriveAspUserLeaf, verifySelectiveDisclosure };
```
Fantoni0: Do we need DisclosureRequest and Disclosure Receipt here?

Fantoni0 (changes requested):
Left some more comments @v0-e

sdk/web/js/index.js
v0-e: Only DisclosureRequest. Also removed PrivatePool and PoolExecuteResult since they are not needed.

sdk/web/src/models/disclosure.rs
v0-e: It is intentional, to make amounts human-readable.

sdk/web/src/client/pool.rs
v0-e: Yep, fixed that :+1:

Fantoni0 (dismissed):

Fantoni0 (approved):

PR #543, opened 2026-09-02
remove relaxed ASP insertion and update_admin panic

### LINKED ISSUE #370: Remove relaxation of admin controls (`set_admin_insert_only`) Fix update_admin panicking

dkgoutham:
## What changed

Two authorization defects in the ASP contracts, plus the SDK and admin-page changes that
follow from removing the first one.

**The allowlist could be written by anyone.** `asp-membership` exposed
`set_admin_insert_only`, which wrote a boolean `AdminInsertOnly` into persistent storage.
`insert_leaf` read that flag and, when it was `false`, skipped `require_auth` entirely:

```rust
let admin_only: bool = store.get(&DataKey::AdminInsertOnly).unwrap_or(true);
if admin_only {
    let admin: Address = store.get(&DataKey::Admin).ok_or(Error::NotInitialized)?;
    admin.require_auth();
}
```

Any key could then append itself to the membership tree. Because a pool with `ALLOWLIST_BIT`
set proves deposits against that tree's root, every such pool would accept deposits from
arbitrary accounts for as long as the flag stayed off. One admin call opened it, and nothing
closed it automatically.

The flag is removed rather than pinned to `true`. The setter, the `DataKey` variant, and the
constructor write are all gone, because a stored boolean that gates an authorization check is
one setter away from being a bypass again. `insert_leaf` now has a single path with no branch:
load `Admin`, return `NotInitialized` if it is absent, then `require_auth`.

**Admin rotation panicked instead of returning an error.** `soroban_utils::update_admin`
loaded the stored admin with `.expect("admin not initialized")`, so a contract whose `Admin`
entry was missing aborted the invocation with a string panic rather than a contract error.
Three of the four callers papered over this with a `has(&DataKey::Admin)` pre-check, a second
storage read whose exact sequence had already drifted between the copies. The fourth,
`asp-non-membership::update_admin`, had no pre-check and returned `()`, leaving it the one
rotation entry point that still panicked. The helper now returns `Result<(), AdminError>`,
every caller maps that to its own `Error::NotInitialized`, and the pre-checks are deleted, so
there is one load-and-require sequence to audit instead of four variants of one.

Where to look: `contracts/asp-membership/src/lib.rs` carries the bypass removal and is the
highest-risk file in the diff. `contracts/soroban-utils/src/utils.rs` carries the helper
change that the other three contracts depend on. The remaining contract changes are
mechanical, and the SDK, app, and README changes only drop the field and the UI that the
removed flag fed.

Test coverage grew from 117 to 128 across the four contract crates. Every privileged entry
point in all four now has a test that it refuses an unauthorized caller, which none of
`asp-non-membership`'s had before, and each ASP has a test that a rotated-out admin loses
insert rights. Each of those tests was checked by deleting the `require_auth` it covers and
confirming the test fails.

## Breaking changes

1. `asp-membership::set_admin_insert_only` is gone from the contract interface, and the
   `AdminInsertOnly` ledger entry is no longer written. A client calling that method must drop
   the call.
2. `asp-non-membership::update_admin` returns `Result<(), Error>` instead of `()`. Clients
   generated from the old contract spec need regenerating; none are checked in.
3. `types::AspMembership` no longer carries `admin_insert_only`. This is semver-major for the
   SDK crates. The `check` job in `release.yml` runs `cargo-semver-checks` only when the
   version line in `sdk/native/Cargo.toml` changes, so CI does not flag it on this branch; the
   next version bump will.

Removing `AdminInsertOnly` from the `DataKey` enum does not move the other keys. Soroban
encodes a `#[contracttype]` unit variant by its symbol name rather than by ordinal, so `Root`,
`Levels`, `NextIndex`, and `Admin` keep the same ledger keys on an existing deployment. A
stale `AdminInsertOnly` entry left behind by an already-deployed contract is simply never read
again. `test_legacy_admin_insert_only_flag_is_ignored` writes such an entry under the old key
shape and confirms `insert_leaf` still demands admin authorization.

## Design notes

`soroban_utils::update_admin` keeps a `# Panics` section even though it now returns a
`Result`. `require_auth` still traps when the stored admin does not authorize the call, and
that is worth stating on a shared helper four contracts route their rotation through.

The four call sites destructure the error rather than discarding it:

```rust
.map_err(|soroban_utils::AdminError::NotInitialized| Error::NotInitialized)
```

`AdminError` has one variant today, so this is irrefutable and costs nothing. Adding a second
variant later fails to compile at all four sites instead of silently reporting the new case as
`NotInitialized` on an authorization path.

Closes #370

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,11 +57,6 @@ This is [the administrative control panel](https://nethermindeth.github.io/stell
 
 This provides **illicit activity safeguards** while maintaining user privacy. The ASP membership trees work with the zero-knowledge proofs to prove that deposits either belong to approved accounts or don't belong to blocked accounts—without compromising privacy.
 
-The admin has the option of toggling the "Admin-Only Leaf Insert", It's enabled by default which restricts only the admin to insert membership leaves but when disabled by the admin, anyone can insert membership leaves.
-
-> [!WARNING]
-> Disabling "Admin-Only Leaf Insert" removes the access-control safeguard on the ASP membership tree. Any party will be able to add themselves (or others) to the approved set without admin approval, bypassing the intended illicit-activity safeguards. Only disable this in a controlled demo or testing environment—never in production.
-
 #### Zero-Knowledge Circuits
 
 The main transaction circuit proves:
--- a/app/admin.html
+++ b/app/admin.html
@@ -128,33 +128,6 @@
                         </div>
                     </div>
                 </div>
-
-                <!-- Admin Insert Only Toggle -->
-                <div class="rounded-2xl border border-white/8 bg-ink-900/70 p-5 mt-4">
-                    <div class="flex flex-col sm:flex-row sm:items-center justify-between gap-4">
-                        <div>
-                            <p class="text-[11px] font-semibold uppercase tracking-[0.24em] text-slate-500">Admin-Only Leaf Insert</p>
-                            <p class="mt-1 text-xs text-slate-400">When enabled, only the admin can insert membership leaves.</p>
-                        </div>
-                        <div class="flex items-center gap-4 flex-shrink-0">
-                            <span id="adminInsertOnlyStatus" class="font-mono text-sm text-slate-400">--</span>
-                            <button
-                                type="button"
-                                id="toggleAdminInsertOnlyBtn"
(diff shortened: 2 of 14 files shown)


Fantoni0 (approved):
Great @dkgoutham ! LGTM!

v0-e (approved):

Issue #397, opened 2026-07-16
Support multiple pools for the same asset


Fantoni0:
Right now, we can deploy multiple pools for different assets with ease.
However, if we want to support more than 1 pool per asset the UI is not able to properly render them because of json-rpc error:

<img width="583" height="156" alt="Image" src="https://github.com/user-attachments/assets/db4df507-d4f8-4daf-8fac-51ae09edc7e3" />

Example:
You want to deploy an XLM pool with policy A, and a second XLM pool with a policy B.

DarkLord017:
can i take this up @Fantoni0

Fantoni0:
Sure @DarkLord017 go ahead

v0-e:
One issue is that contract ID limit for `getEvents` is 5 per filter (as the attached screenshot reports). And we should fix that.
Maybe multiple pools for same asset is not the cause itself. Most likely this was tested with a deployment with also ASP/non-ASP contracts? That could trigger the error since that is more contracts.

Fantoni0:
Also, if we support mulitple pools for the same assets. We need to differentiate them in the UI. It would be great to include some tags in the overview to identify the pool configuration. If it has:
- Whitelist 
- Blocklist
- Both
- GVK?

v0-e:
:+1: #384 is the relevant issue for that
Likely this issue is closed with #544. I'm guessing the issue was number of contracts rather than multiple pools for the same asset
here.

Fantoni0:
Also, we need to differentiate pools on the Advanced tab. Otherwise:

<img width="515" height="185" alt="Image" src="https://github.com/user-attachments/assets/f67bb2c7-204c-4f4f-ab13-ef44a76a439f" />

Fantoni0:
Good call, @v0-e , forgot about #384. Moved the comment and closed this one.

PR #544, opened 2026-09-02
Increase number of fetched contract IDs in `getEvents`

v0-e:
Increases the number of fetched contract IDs from 5 to 25 in `getEvents` by splitting them into different filters (up to 5, 5*5=25).

Related with #397.


Fantoni0 (approved):
LGTM!

PR #545, opened 2026-09-03
Clear unowned Chrome singleton locks before every profile launch

cupicmarko:
Chrome writes `SingletonLock` — a symlink to `<hostname>-<pid>` — plus
`SingletonSocket` and `SingletonCookie` at the root of a user-data-dir while it
runs, and removes them on a clean exit. A killed run leaves them behind. The
next launch does not fail on them: it decides the profile is in use elsewhere
and quietly switches to a throwaway one. The extension loads but has no wallet,
so the app reports Freighter as not detected while the provisioned profile sits
intact on disk.

The profile this reaches is the live `e2e-freighter/.chrome-profile`.
`provision.sh --verify`, and plain `provision.sh` when a snapshot already
exists, both run `provision.mjs --verify`, which launches Chrome against that
directory directly (`scripts/provision.sh:69-74`, `scripts/provision.mjs:264`).
Nothing on that path removes a lock. The rebuild path is not affected — it
starts with `rm -rf "$PROFILE_DIR"` — and a restored profile cannot carry one,
because the snapshot tar already excludes all three names.

## Change

`src/chrome-locks.mjs` exports `clearSingletonLocks(userDataDir)`, and
`launch()` in `src/runner.mjs` calls it before
`chromium.launchPersistentContext`.

The presence of a lock does not prove its owner is gone, so removal is
conditional. `SingletonLock`'s target is read and compared against this host
and the live process table:

- target names a dead pid, or a host that is not this one → all three files are
  removed, and `launch()` warns with the names it removed;
- target names a live process on this host → throws, and removes nothing;
- target cannot be read as `<hostname>-<pid>` — a regular file, a directory, an
  unparseable link → throws, and removes nothing.

Refusing is the point of the last two. Two browsers sharing one profile
corrupts it silently; an aborted launch names the pid and the path that
stopped it. The hostname and liveness probe are injectable so the refusal
paths are testable without a browser; the default probe is `process.kill(pid, 0)`,
treating `EPERM` as alive.

`launch()` is the only place this package starts a browser: `provision.mjs` uses
it for both the `--verify` launch and the headed provisioning launch, and
`runner.mjs` uses it for every test run. Putting the cleanup there covers the
live profile and the restored temp copy without either script needing to know
about locks.

`tests/unit/chrome-locks.test.mjs` covers both halves. Removal: a dead pid, and
another host's live pid, both clear all three files; a socket and cookie with no
lock claim no owner; unrelated profile contents survive; an absent profile is a
no-op. Refusal: a live pid on this host, a lock that is a regular file, a lock
that is a directory, and an unparseable target each throw with nothing removed —
including one case that plants this test process's own pid, so the real liveness
probe is exercised rather than a stub.

### DIFF
--- a/e2e-freighter/src/chrome-locks.mjs
+++ b/e2e-freighter/src/chrome-locks.mjs
@@ -1,0 +1,68 @@
+// Remove Chrome's single-instance locks from a user-data-dir, but only when
+// nothing still holds them.
+//
+// Chrome writes SingletonLock — a symlink to "<hostname>-<pid>" — plus
+// SingletonSocket and SingletonCookie while a profile is open, and removes
+// them on a clean exit. A killed run leaves them behind, and the next launch
+// reads them as "open elsewhere" and quietly switches to a throwaway profile:
+// the extension loads with no wallet.
+//
+// The rule enforced here: locks are removed only when the target names a dead
+// pid or a different host. A live owner, or a target that cannot be read as
+// "<hostname>-<pid>", throws — an aborted launch is diagnosable, two browsers
+// on one profile is not.
+import { lstatSync, readlinkSync, rmSync } from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+const SINGLETON_NAMES = ['SingletonLock', 'SingletonSocket', 'SingletonCookie'];
+
+function defaultIsProcessAlive(pid) {
+  try {
+    process.kill(pid, 0);
+    return true;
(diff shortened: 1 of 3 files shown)


v0-e (dismissed):
lgtm!

Fantoni0 (dismissed):
Once linter issue is fixed, LGTM. Great @cupicmarko

v0-e (approved):

PR #546, opened 2026-09-03
Scope the archive endpoint and logging posture per account

cupicmarko:
`app/js/app-storage.js` stored two settings under one key each, shared by every
account on the profile:

- `bootnode_config` — the archive endpoint used to backfill history the main RPC
  can no longer serve. Connecting a second account sent its historical sync to
  whichever endpoint the first account had chosen, without asking.
- `telemetry_config` — the log level and `revealSensitive`, which un-redacts
  amounts and addresses in the telemetry ring buffer and in UI toasts. A second
  account started with the first account's posture already armed.

`explorer` was stored the same way, but it is public infrastructure and
identical for everyone.

## Change

`settingKey(key, address)` namespaces an account-scoped setting as
`key::address`, and throws if the address is missing rather than resolving to
the bare key — the silent fallback is how the inheritance happened.
`ACCOUNT_SCOPED_SETTINGS` holds the two; `explorer` is not in it and resolves to
one key for every account.

`getSetting`/`setSetting` take the address and route through it.
`getBootnodeConfig`/`setBootnodeConfig` require one, and
`getTelemetryConfig`/`setTelemetryConfig` are added alongside them. Every call
site passes the account it is acting for: `bootnodeCheck` and the connect
sequence in `app/js/ui/navigation.js`, the wizard's plan and its retention write
in `app/js/ui/onboarding-wizard.js`, and `initializeRuntime` in
`app/js/wasm-facade.js`, which resolves the stored endpoint when the caller does
not pass one explicitly.

`Wallet.disconnect()` clears the two settings from `App.state` instead of
leaving them for the next connect to overwrite — the settings drawer is
reachable while connecting, so a stale value is visible to the next account. It
also resets the telemetry posture in the wasm, which is a process global rather
than part of the disposed client and would otherwise stay armed. `saveSettings`
refuses to write without a connected account.

Legacy unscoped values are not migrated: they belonged to whichever account was
active when they were written, and adopting them for a different account is the
defect. One caller has no account to scope to — `app/js/admin.js` probes for an
endpoint from `init()`, before any wallet is connected — so
`getStoredBootnodeUrl` reads the unscoped key when given no address. Nothing
writes that key any more, so the fallback only ever sees data from before this
change, and an account that simply has no endpoint of its own does not reach it.

The wizard showed the retention step whenever no `bootnode_config` record
existed. Per-account, an already-onboarded account has no such record and never
will, so the step would return on every connect. It is now shown when the
account has not completed onboarding (disclaimer accepted and keys derived); a
genuine sync gap still forces it through `bootnodeRequired`, which is where an
existing account is asked again at the point the endpoint is actually needed.

`e2e-freighter/tests/unit/settings-scope.test.mjs` pins the rule: both settings
scoped and `explorer` not, two accounts never sharing a scoped key, a scoped
read without an address throwing rather than silently unscoping, the bare key
never produced for a scoped setting, and the accountless fallback reading the
old value while a scoped read does not.

### DIFF
--- a/app/js/admin.js
+++ b/app/js/admin.js
@@ -200,11 +200,11 @@ async function ensureCryptoReady() {
     try {
       const storage = await ensureStorage();
       if (await bootnodeRequired(sorobanRpcUrl)) {
-        if (!(await storage.getStoredBootnodeUrl())) {
+        if (!(await storage.getStoredBootnodeUrl(state.address))) {
           throw new Error('RPC_SYNC_GAP: bootnode required');
         }
       }
-      await initializeRuntime(sorobanRpcUrl);
+      await initializeRuntime(sorobanRpcUrl, { address: state.address });
       await client().backgroundSync();
     } catch (e) {
       if (isDbLockedError(e?.message)) showDbLockedModal(e.message);
--- a/app/js/app-storage.js
+++ b/app/js/app-storage.js
@@ -5,6 +5,36 @@
 
 const SETTING_EXPLORER = 'explorer';
 const SETTING_BOOTNODE_CONFIG = 'bootnode_config';
+const SETTING_TELEMETRY_CONFIG = 'telemetry_config';
+
+// A setting is account-scoped if its value can reveal something about the
+// account it is applied under; global if it is a presentation preference that
+// is the same for everyone.
+//
+// bootnode_config: its operator sees which account is syncing, so sharing one
+// endpoint across accounts links them.
+// telemetry_config: revealSensitive un-redacts this account's amounts and
+// addresses, and level decides how much of its activity is kept.
+// explorer: public infrastructure, identical for every account.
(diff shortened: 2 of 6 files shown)


v0-e (commented):
Perhaps is best to avoid the per-account bootnode config. In principle it is OK, but mixing bootnodes against the same indexed data (shared by all accounts) is something that I feel is a bit risky. Both client-side due to pagination logic and server-side due to potentially different data being provided.

I assume here that the reason for this was originated by the onboarding flow being specific to each account. But perhaps is best to share some settings among the accounts.

cupicmarko:
> I assume here that the reason for this was originated by the onboarding flow being specific to each account. But perhaps is best to share some settings among the accounts.

@v0-e Ok, I agree, I'll review some other strategies, and close this PR in the meantime, not to take your time on it until it's ready.

PR #547, opened 2026-09-03
Wrap the account address in the missing-keys errors

cupicmarko:
`Sensitive` wraps addresses and amounts at every logging site in `sdk/native`.
Three error constructors were the exception:

- `sdk/native/src/transact.rs:218` — prints the account address when no privacy
  keys are stored. Runs in the browser storage worker, so it reaches a UI toast
  and the telemetry ring buffer.
- `sdk/native/src/types/amounts.rs:122` — prints the amount on an out-of-range
  conversion, reaching the same two sinks.
- `sdk/native/src/storage/mod.rs:42` — a byte-identical copy of the missing-keys
  message for the `LocalStorage` path, reaching CLI output and logs.

## Change

All three are wrapped in `Sensitive`, so the value renders as `<redacted>` unless
the reveal flag is set in a debug build. The CLI operator already knows which
address failed — these commands take one account from config and print it — and
the flag restores the value when debugging.

`REVEAL_SENSITIVE` is a process global, so a redaction assertion in any module
races the tests in `types/logging.rs` that toggle it. The mutex serialising those
tests moves beside the flag as `lock_reveal_flag()` and is shared. It ignores
poisoning: each caller sets the flag it needs, and a cascading poison error would
hide the real failure. All `#[cfg(test)]`, so nothing ships.

One test per site asserts against the rendered error string that the raw value is
absent and `<redacted>` is present.

### DIFF
--- a/sdk/native/src/storage/mod.rs
+++ b/sdk/native/src/storage/mod.rs
@@ -40,8 +40,10 @@ pub(crate) fn map_user_keys(
         .get_user_keys(user_address)
         .map_err(|e| Error::Other(e.to_string()))?
         .ok_or_else(|| {
+            // Escapes to CLI output and logs.
             Error::Other(format!(
-                "address {user_address} should generate privacy keys and ASP secret first"
+                "address {} should generate privacy keys and ASP secret first",
+                crate::types::Sensitive(user_address)
             ))
         })
 }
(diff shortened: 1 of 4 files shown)


v0-e (commented):
Looks good!
- AI also flagged `sdk/native/src/state/storage.rs:257`,  raw address leaking through `.context(format!(...))`
- I think the lock system is fine, but to make serialized tests, the `#[serial]` annotation from the `serial_test` crate is cool for this. Feel free to ignore

Fantoni0 (dismissed):
LGTM! 🚀 
On the linter issue, I had ro run `rustup update` yesterday to stay up to date with CI version (thanks @v0-e )

v0-e (approved):

PR #548, opened 2026-09-03
Require both account identities at session open and refuse a divergent pair

cupicmarko:
`AccountOptions.signerAddress` is accepted but not honoured. `Client::account`
in `sdk/web/src/client/mod.rs` builds a `WalletSigner` from it, so the wallet is
asked to sign as that account — but `open_native_account` passes only
`userAddress` to the native client, and `sdk/native/src/client.rs` then discards
whatever was supplied and sets `signer_address = SignerAddress::new(user_address)`.

A caller passing `userAddress: A` with `signerAddress: B` therefore gets a
session where the wallet signs as B while the envelope source, the pool
contract's `sender`, the sequence-number lookup and the auth entries are all
built for A. The `Account.signerAddress` getter reports A, contradicting its own
doc comment, so nothing on screen or in the API reveals which account is
actually signing.

The first session for an account compounds it. `Client::account` derives privacy
keys when none exist: it signs `KEY_DERIVATION_MESSAGE` with `signerAddress` and
files the result under `userAddress`. Those keys are gated by a
`user_keys_exist` check, so they are never re-derived — one divergent open
permanently files A's privacy keys as a function of B's signature, and every
later session for A reuses them.

The two identities already have distinct types, and
`sdk/native/src/types/address.rs` states in its own doc comment that there is
deliberately no conversion between them. There is: `address_newtype!` implements
`From<String>` and `From<&str>` for both. With `Client::account` taking
`impl Into<NoteOwnerAddress>`, an owner-derived string crossed into either type
through an `.into()` naming neither, which is how the two came to be conflated.

## Change

`Client::account` and its blocking mirror take `user_address: NoteOwnerAddress`
and `signer_address: SignerAddress` as separate arguments. A pair naming two
different accounts is refused with `Error::SignerIsNotNoteOwner`, which carries
both addresses as structured fields.

Refusing rather than resolving is deliberate: the two transaction paths disagree
about which identity drives them. `prepare_pool_transact` takes the signer as
its `source_account`, so transact builds the envelope, the contract's `sender`
and the sequence-number lookup from the signing account; `prepare_register` uses
the note owner for both the registry key and the transaction source, and says in
its own comment that which it should be when the two differ is unsettled. A
divergent pair would file a registration under one account and spend under
another with no rule saying which is right, so it fails instead of silently
picking one.

`sdk/web/src/client/mod.rs` performs the same check as soon as both addresses
are resolved — before `WalletSigner::new`, before the `user_keys_exist` read,
before the wallet is prompted and before `derive_save_user_keys` runs. It raises
the native `Error` variant rather than a parallel message, so the two cannot
drift. The native check remains the authoritative one.

Both addresses are wrapped in `Sensitive` when the error renders. Addresses are
Tier-1 in this crate and this message reaches a UI toast, the telemetry ring
buffer and CLI logs; the structured fields stay in the clear for code that needs
to distinguish the pair.

`From<String>` and `From<&str>` are removed from `address_newtype!`. `new` is
now the only constructor, so every crossing writes its destination type down at
the site where it happens. Call sites in the CLI, the examples, the SDK
fixtures and the browser client construct both wrappers explicitly from the one
address they hold; behaviour there is unchanged, since all four already used a
single account for both roles.

`WalletSigner` gains a `signer_address()` accessor so `open_native_account`
reads the address the wallet will actually be asked to sign with, rather than
re-reading `AccountOptions`.

The `Account.signerAddress` getter doc no longer claims it can differ from
`userAddress`, and `AccountOptions.signerAddress` in `options.d.ts` records that
a differing value is refused before the wallet is prompted.

Tests cover the guard directly (accepted, refused, and an exactness check over
case, leading whitespace and empty), through the public `Client::account` on
both the accept and refuse paths, and on the browser guard. The type invariant
is held by const assertions that resolve during type-checking — a reinstated
`From` impl fails to compile rather than reporting a test failure — plus a
`compile_fail` doctest showing the `.into()` crossing the split exists to
prevent, and a passing doctest showing the explicit construction that replaces
it.

### DIFF
--- a/sdk/native/src/blocking/client.rs
+++ b/sdk/native/src/blocking/client.rs
@@ -1,6 +1,8 @@
 //! Sync wrapper around [`crate::Client`] via a shared Tokio runtime.
 
-use crate::types::{ContractConfig, NoteOwnerAddress, OperationalFeedItem, RecipientLookup};
+use crate::types::{
+    ContractConfig, NoteOwnerAddress, OperationalFeedItem, RecipientLookup, SignerAddress,
+};
 
 use crate::{
     BackgroundSync, Error, Handle, Prover, Signer, chain::StateFetcher,
@@ -70,14 +72,19 @@ impl Client {
         block_on(self.inner.recipient_lookup(address))
     }
 
+    /// Blocking mirror of [`crate::Client::account`]; the same
+    /// owner-must-equal-signer rule applies.
     pub fn account(
         &self,
-        user_address: impl Into<NoteOwnerAddress>,
+        user_address: NoteOwnerAddress,
+        signer_address: SignerAddress,
         signer: Handle<dyn Signer>,
     ) -> Result<Account, Error> {
-        Ok(Account::from_inner(
-            self.inner.account(user_address, signer)?,
-        ))
+        Ok(Account::from_inner(self.inner.account(
+            user_address,
+            signer_address,
+            signer,
+        )?))
     }
 
     pub fn state_fetcher(&self) -> Result<StateFetcher, Error> {
--- a/app/ARCHITECTURE.md
+++ b/app/ARCHITECTURE.md
@@ -25,7 +25,7 @@ The web SDK runs Rust on the main thread via WASM, with blocking work offloaded 
 ### Lifecycle
 
 ```
-init() → Storage.open() → bootnodeRequired() → Client.new() → backgroundSync() → client.account(signer) → account.pool() → PrivatePool ops
+init() → Storage.open() → bootnodeRequired() → Client.new() → backgroundSync() → client.account(options, signer) → account.pool() → PrivatePool ops
 ```
 
 The app wraps this in `wasm-facade.js` and `ui/pool.js`: `bootnodeRequired` → `initializeRuntime` → `client().backgroundSync` → `client().openAccount` → `account().pool()` via `createAppPool()` / `ensureAppPool()`.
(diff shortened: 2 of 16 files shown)


sdk/native/src/blocking/client.rs
```diff
@@ -70,14 +72,19 @@ impl Client {
         block_on(self.inner.recipient_lookup(address))
     }
 
+    /// Blocking mirror of [`crate::Client::account`]; the same
+    /// owner-must-equal-signer rule applies.
     pub fn account(
         &self,
-        user_address: impl Into<NoteOwnerAddress>,
+        user_address: NoteOwnerAddress,
+        signer_address: SignerAddress,
```
v0-e, line 80: I get the separation, but if they are enforced to be the same, then we are just adding overhead to the user.
Any way to merge the address args? At least separate them inside the fn, not in the fn signature?

cupicmarko: Made a small compromise to address your comment because it is correct, but we will need distinction in future changes, we now have `account_for` so it's simplified for the user, but we're still able to separate them when needed.

cupicmarko: I don't want to remove the previous comment, it should be there for documenting my decision, but I reverted that because it's against my intention for next PRs: 
Keeping both arguments: cli/src/session.rs is about to take a --sign-as flag.

v0-e (approved):

PR #549, opened 2026-09-03
Fix onboarding issue in freighter setup

Fantoni0:
Fixes the current issue in CI related to the setup of Freighter E2E tests.

The onboarding checked for bootnodeRequired and treated it as an onboarding gate. But the actual probing to check if we are behind the RPC retention window stays true (until we are finally synced). So every time re re-connected, the wizard was prompted again (even if it was saved and we were already catching up)


v0-e (approved):
lgtm!

PR #550, opened 2026-09-03
fix(pool): bind ext_data_hash to pool and token identity

### LINKED ISSUE #539: Bind transaction proofs to pool contract and token domain
Current transaction proofs bind values such as root, publicAmount, extDataHash, nullifiers and commitments, but they do not bind the proof to the pool contract or token/asset.

ExtData also contains no pool or asset identifier, so extDataHash does not provide that separation either.

A local two-pool regression test confirms that two distinct pool contracts using distinct tokens, the same verifier, and the same tree depth are indistinguishable through all pre-verifier transaction checks: both start from the same genesis root, and the same proof/ExtData pair reaches the verifier boundary in both contracts without UnknownRoot, WrongExtHash, or WrongExtAmount.

This matters because the deployment tooling intentionally reuses the same verifier/VK across pools with compatible policy flags, including pools for different assets.

The agreed fix is to bind extDataHash to both the current pool contract address and the configured token contract address, mirrored consistently on-chain and in the SDK.

The intended invariant is that a proof/ExtData pair created for one pool/token domain must not remain valid for another pool/token domain, even when both deployments share the same verifier, circuit, tree depth, and genesis root.

Related: #397, #371.

No direct fund-theft exploit or successful cross-pool proof reuse is claimed here. The issue is the absence of a protocol-level guarantee that a proof is scoped to the pool/token domain it was generated for.

AngeYobo:
## 🚀 What’s this PR do?

Bind transaction proofs to the current pool contract and configured token domain.

`extDataHash` now includes the canonical pool contract address and pool token address, preventing a proof/ExtData pair created for one pool/token domain from remaining valid against another compatible deployment.

The cross-pool regression now rejects with `WrongExtHash` before verifier invocation.

---

### 📎 Related issues (optional)

Closes #539

---

## 🧠 Context

On-chain, the domain is derived from `env.current_contract_address()` and the canonical token stored by the pool.

The native SDK mirrors the same Soroban/XDR encoding when computing `extDataHash`.

There is no legacy fallback, caller-supplied domain override, circuit change, VK change, trusted setup change, or network-domain binding in this PR.

Validation completed locally:

- `cargo test -p pool -p pool-gvk -p pool-core`: 74/74
- `cargo test -p stellar-private-payments --lib`: 297/297
- `cargo test -p e2e-tests --lib`: 10 passed, 1 ignored
- `cargo test -p sdk-tests --lib`: 17/17
- `cargo test --workspace --verbose`: pass
- `cargo clippy --all-targets -- -D warnings`
- `cargo +nightly fmt --all -- --check --config-path rustfmt.toml`
- `cargo shear`
- `cargo deny check`
- `typos`
- JS license audit
- Soroban contract builds for `asp-membership`, `asp-non-membership`, `circom-groth16-verifier`, and `pool`

`cargo sort --workspace -g --check` also fails on current `upstream/main`; this PR does not modify any `Cargo.toml`.

I ran `.githooks/pre-push` locally. It stops at `cargo sort --workspace -g --check`, which also fails on current `upstream/main`. This PR does not modify any `Cargo.toml`.

The remaining pre-push gates were run individually and passed, including workspace tests, clippy, fmt, cargo shear, cargo deny, typos, JS license audit, and Soroban contract builds.

---

## ✅ Required Checklist


contracts/pool-core/src/ext_data.rs
```diff
@@ -25,25 +26,141 @@ pub struct ExtData {
...
-/// Hash external data using Keccak256
+/// `ExtData` plus the deployment domain it is bound to.
 ///
-/// Serializes the external data to XDR, hashes it with Keccak256,
-/// and reduces the result modulo the BN256 field size.
+/// Never constructed from caller input: `pool` and `token` are read by
+/// `hash_ext_data` from `env.current_contract_address()` and the caller's own
+/// canonical token configuration, never from `ExtData` or any other
+/// caller-supplied value. This is what makes the resulting hash, and so the
+/// proof, specific to one pool contract and one token.
```
Fantoni0, line 35: Using AI for documentation is fine. But let's try to keep it constrained. We do not need multiple paragraphs for structs.

This applies to the whole PR.

contracts/pool-gvk/src/test.rs
```diff
@@ -1573,23 +1581,17 @@ fn build_gvk_transact(
     let (vk_bytes, real_proof) = groth16_fixture_for(&env, &values);
     proof.proof = real_proof;
 
+    // Swap in the fixture-matched verifier now that it exists. `pool_id`
```
Fantoni0, line 1584: Rather than poking the storage manually. You can use Soroban's Env::register_at(&address, contract, args). So you reserve the pool address, build the hash, and then register the real pool at the address.

Fantoni0 (changes requested):
Thanks for the contribution @AngeYobo 
Left some comments, but looks pretty good! 👐

AngeYobo:
Thanks @Fantoni0, makes sense.

I’ll tighten the comments and switch the test to Env::register_at. 🙌

AngeYobo:
Addressed. Now reserving the pool address first and registering the real pool with Env::register_at. Thanks.

Fantoni0 (approved):

Fantoni0:
Was going to merge, but last commit is not signed @AngeYobo

AngeYobo:
Ah yes, I made that commit through GitHub Codebase while traveling. I’ll
sign it properly as soon as I’m back.

Thanks @Fantoni0

PR #553, opened 2026-09-03
• fix(web): isolate concurrent E2E test accounts

cupicmarko:
Provision four ephemeral accounts and assign a dedicated source account to
  each mutating browser test. Select the matching signer credentials and isolate
  progress-event capture by flow to prevent concurrent tests from interfering.

### DIFF
--- a/.github/workflows/e2e-webclient.yml
+++ b/.github/workflows/e2e-webclient.yml
@@ -73,18 +73,24 @@ jobs:
         env:
           E2E_BOOTNODE_URL: https://bootnode.dev-nethermind.xyz
         run: |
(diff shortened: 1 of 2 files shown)


Fantoni0 (approved):

v0-e (commented):
Just a question, do you think we need to provision the C/D accounts, since tests are now sequential?

cupicmarko:
> Just a question, do you think we need to provision the C/D accounts, since tests are now sequential?

@v0-e, I would keep provisioning all accounts, for loop in the workflow makes the execution sequential, but running tests with one command locally with A and B only could cause the same.

v0-e (approved):
I see :+1:

PR #560, opened 2026-09-04
Fix bootnode rate limit RPS period

v0-e:
Tower governor `per_second` is the opposite of what we wanted. Using `period` now instead.


Fantoni0 (approved):
LGTM.period()

Issue #394, opened 2026-07-15
Enrich SDK main error type


v0-e:
Current SDK errors default a lot to `Other` (its `anyhow` variant), making a bit difficult error handling by the user.

PR #561, opened 2026-09-04
SDK Other error as anyhow

v0-e:
Initial efforts to improve SDK error handling (#394).
While the goal is to decrease the use of strings and anyhow use throughout the library, and type most instances, the conversion from `String` to `anyhow::Error` in this PR makes a small incremental improvement: anyhow can at least preserve the full error chain.

Further changes to come in the stack.

### DIFF
--- a/sdk/native/src/core/mod.rs
+++ b/sdk/native/src/core/mod.rs
@@ -52,7 +52,7 @@ impl PoolCore {
             self.config.pool_contract_id.clone(),
             SpendTarget::transfer(note_public_key, encryption_public_key),
         )?;
-        PreparedTransactionPlan::from_session(session).map_err(Error::from)
+        Ok(PreparedTransactionPlan::from_session(session)?)
     }
 
     pub fn prepare_withdraw(
@@ -70,7 +70,7 @@ impl PoolCore {
             self.config.pool_contract_id.clone(),
             SpendTarget::withdraw(recipient.into()),
         )?;
-        PreparedTransactionPlan::from_session(session).map_err(Error::from)
+        Ok(PreparedTransactionPlan::from_session(session)?)
     }
 
     pub fn estimate(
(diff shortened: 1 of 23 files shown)


sdk/native/src/core/mod.rs
```diff
@@ -90,8 +90,9 @@ impl PoolCore {
         enc_pub: EncryptionPublicKey,
         amount: NoteAmount,
     ) -> Result<Transact, Error> {
-        let ext_amount = ExtAmount::try_from(amount)
-            .map_err(|_| Error::Other("deposit amount exceeds ext_amount range".into()))?;
+        let ext_amount = ExtAmount::try_from(amount).map_err(|_| {
+            Error::Other(anyhow::anyhow!("deposit amount exceeds ext_amount range"))
```
Fantoni0: AI complains about this not maintaining the whole chain of errors.  Derives the display from anyhow:Error, which by default only prints the outermost context message.

sdk/native/src/sync.rs
```diff
@@ -336,12 +337,9 @@ async fn apply_bootnode_handoff<S: Storage>(
...
         .save_sync_progress(metadata, false)
         .await
-        .map_err(|e| Error::Other(format!("handoff sync progress: {e:#}")))?;
+        .context("handoff sync progress")?;
     storage.clear_indexing_cursors().await?;
-    storage
-        .clamp_last_fully_indexed_ledger(from_ledger)
-        .await
-        .map_err(|e| Error::Other(format!("handoff clamp fully indexed: {e:#}")))?;
+    storage.clamp_last_fully_indexed_ledger(from_ledger).await?;
```
Fantoni0: Minor, but we lost the context mesasge of "handoff clamp fully indexed"

Fantoni0 (dismissed):
Left some comments. Feel free to ignore if you don't agree

v0-e:
Addressed the comments :+1:

Fantoni0 (approved):
LGTM!

PR #562, opened 2026-09-04
Extend storage TTL on every contract call

dkgoutham:
Soroban charges rent on ledger entries. An entry that goes unrenewed past its expiry is
archived: the host can no longer read it, and any call whose footprint touches it fails until
somebody restores it. Nothing in this tree renewed anything, so every contract was running
down a clock nobody was watching.

The practical shape of the bug: a pool with no traffic for long enough had its Merkle state
archived. The next `transact` then failed reading entries that still existed on chain but
were no longer live, and no amount of retrying fixed it.

## The `ttl` module in `soroban-utils`

Two constants and three helpers, so every contract renews on the same schedule instead of
each one inventing its own.

- `THRESHOLD` is 30 days of ledgers, `EXTEND_TO` is 180. A renewal is skipped entirely while
  an entry still has more than `THRESHOLD` left, so the cost lands only on calls that would
  otherwise let something lapse. A busy pool pays almost nothing; a pool waking up after
  months pays once.
- `bump_instance` renews the calling contract's own instance entry and its code.
- `bump_entry` renews one persistent key.
- `bump_dependency` renews a linked contract's instance and code together, through
  `deployer().extend_ttl`. Pools use it on the verifier and the two ASP contracts they call,
  because a live pool with an archived verifier is just as stuck.

## Applied across the contracts

Every entry point in `pool`, `pool-gvk`, `asp-membership`, `asp-non-membership`, and
`pool-core` renews its own instance plus the entries it reads or writes on that call.

The deliberate limit is that a call renews **only what it touches**. Renewing everything a
pool owns would mean a `transact` bumping all ninety root-history slots and every level's zero
hash, adding over a hundred entries to its footprint and charging every user for entries their
transaction never read.

The consequence is worth being explicit about: entries that no call touches are not covered
here and stay on their existing expiry. That includes an idle pool's untouched ring slots, old
nullifiers, and the sparse tree's long tail. Reaching those needs something that runs without
a user transaction, which is not in this change.

## Compatibility

No signature changes, no new error variants, no storage layout changes. Deployed contracts
keep working and begin renewing themselves on their next call, with no migration and no
redeploy.

## Tests

Most of the diff. Each contract has a test that advances the ledger until its entries fall
below `THRESHOLD`, calls an entry point, and asserts the entries jump back to `EXTEND_TO`.
One asserts that a call which fails renews nothing, so a reverted transaction cannot be used
to extend state for free.

### DIFF
--- a/contracts/asp-membership/src/lib.rs
+++ b/contracts/asp-membership/src/lib.rs
@@ -8,7 +8,7 @@
 use soroban_sdk::{
     Address, Env, U256, Vec, contract, contracterror, contractevent, contractimpl, contracttype,
 };
-use soroban_utils::{get_zeroes, poseidon2_compress};
+use soroban_utils::{bump_entry, bump_instance, get_zeroes, poseidon2_compress};
 
 /// Storage keys for contract persistent data
 #[contracttype]
@@ -120,6 +120,7 @@ impl ASPMembership {
     /// Returns [`Error::NotInitialized`] if the contract has no admin address
     /// stored.
     pub fn update_admin(env: Env, new_admin: Address) -> Result<(), Error> {
+        bump_instance(&env);
         soroban_utils::update_admin(&env, &DataKey::Admin, &new_admin)
             .map_err(|soroban_utils::AdminError::NotInitialized| Error::NotInitialized)
     }
@@ -134,12 +135,15 @@ impl ASPMembership {
     /// # Returns
     /// The current Merkle root as U256
     ///
-    /// # Panics
-    /// Panics if the contract has not been initialized
+    /// # Errors
+    ///
+    /// Returns [`Error::NotInitialized`] if no root is stored.
     pub fn get_root(env: Env) -> Result<U256, Error> {
+        bump_instance(&env);
         env.storage()
             .persistent()
             .get(&DataKey::Root)
+            .inspect(|_| bump_entry(&env, &DataKey::Root))
             .ok_or(Error::NotInitialized)
     }
 
@@ -156,6 +160,7 @@ impl ASPMembership {
     /// # Returns
     /// The Poseidon2 hash result as U256
     pub fn hash_pair(env: &Env, left: U256, right: U256) -> U256 {
+        bump_instance(env);
         poseidon2_compress(env, left, right)
     }
 
@@ -177,14 +182,18 @@ impl ASPMembership {
     /// [`Error::MerkleTreeFull`] if the tree is at capacity, and
     /// [`Error::Overflow`] if the next leaf index would exceed `u64::MAX`.
     pub fn insert_leaf(env: Env, leaf: U256) -> Result<(), Error> {
+        bump_instance(&env);
         let store = env.storage().persistent();
         let admin: Address = store.get(&DataKey::Admin).ok_or(Error::NotInitialized)?;
+        bump_entry(&env, &DataKey::Admin);
         admin.require_auth();
 
         let levels: u32 = store.get(&DataKey::Levels).ok_or(Error::NotInitialized)?;
+        bump_entry(&env, &DataKey::Levels);
         let actual_index: u64 = store
             .get(&DataKey::NextIndex)
             .ok_or(Error::NotInitialized)?;
+        bump_entry(&env, &DataKey::NextIndex);
         let mut current_index = actual_index;
 
         // Check if tree is full (capacity is 2^levels leaves)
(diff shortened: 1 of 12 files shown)


contracts/pool/src/pool.rs
```diff
@@ -225,6 +225,14 @@ impl PoolContract {
         Ok(())
     }
 
+    /// Extends the TTL of this contract's instance and code entries.
+    ///
+    /// Every public entry point calls this first, so a pool that keeps being
+    /// used never falls behind on rent.
+    fn touch(env: &Env) {
+        bump_instance(env);
+    }
```
v0-e, line 234: I'd remove this wrapper `touch` from pool/pool-gvk. To make it bit more explicit and consistent with the ASP contracts.

contracts/pool-gvk/src/pool_gvk.rs
```diff
@@ -616,7 +643,16 @@ impl PoolGvkContract {
...
+        Self::touch(env);
         sender.require_auth();
+        bump_dependency(env, &Self::get_verifier(env)?);
+        let policy_flags = Self::load_policy_flags(env)?;
+        if policy::requires_membership_proofs(policy_flags) {
+            bump_dependency(env, &Self::get_asp_membership(env)?);
+        }
+        if policy::requires_non_membership_proofs(policy_flags) {
+            bump_dependency(env, &Self::get_asp_non_membership(env)?);
+        }
```
v0-e, line 655: Probably we can remove these bumps since they are a bit redundant? We'll call the ASP contracts down the stack and those will self-bump.

v0-e (commented):
Nice!

contracts/soroban-utils/src/ttl.rs
```diff
@@ -0,0 +1,199 @@
...
+pub const DAY_IN_LEDGERS: u32 = 17_280;
+
+/// Remaining lifetime below which an extension takes effect.
+pub const THRESHOLD: u32 = 30 * DAY_IN_LEDGERS;
+
+/// Lifetime that an extension targets, equal to the network maximum entry TTL.
+///
+/// The host clamps an extension to the network maximum, so this value is safe
+/// on a network whose maximum is lower.
+pub const EXTEND_TO: u32 = 180 * DAY_IN_LEDGERS;
```
Fantoni0, line 26: I think this might be too high. I think the standard is 30 days for extend_to and 7 days for threshold. I think smaller values make it more fair to the average caller

contracts/pool/src/pool.rs
```diff
@@ -431,7 +440,16 @@ impl PoolContract {
...
+        Self::touch(env);
         sender.require_auth();
+        bump_dependency(env, &Self::get_verifier(env)?);
+        let policy_flags = Self::load_policy_flags(env)?;
+        if policy::requires_membership_proofs(policy_flags) {
+            bump_dependency(env, &Self::get_asp_membership(env)?);
+        }
+        if policy::requires_non_membership_proofs(policy_flags) {
+            bump_dependency(env, &Self::get_asp_non_membership(env)?);
+        }
```
Fantoni0, line 452: Right now we have bump_dependency in the user path. This means a user might end up paying the rent for the pool, the verifier, and the ASPs. I think it would be better if each contract pays for its own rent (if possible, I know they are highly coupled). My suggestion would be: explore if it's possible to remove this from user calls, and have a dedicated function that anyone can call (ideally the admin) to extend the TTL. If you don't think this is possible, we can at least use extend_tt__with_limits to keep the costs bounded.

Also applies to pool_gvk

contracts/pool-core/src/merkle_with_history.rs
```diff
@@ -174,7 +178,9 @@ impl MerkleTreeWithHistory {
         // Update the root history index
         root_index = root_index.checked_add(1).ok_or(Error::Overflow)? % ROOT_HISTORY_SIZE;
         // Update the root with the computed hash
-        storage.set(&MerkleDataKey::Root(root_index), &current_hash);
+        let root_key = MerkleDataKey::Root(root_index);
+        storage.set(&root_key, &current_hash);
+        bump_entry(env, &root_key);
```
Fantoni0, line 183: I think we are only bumping partially the ring of roots. Only the recent positions

contracts/pool/src/pool.rs
```diff
@@ -634,8 +664,13 @@ impl PoolContract {
     ///
     /// Returns `true` if the nullifier has been spent, `false` otherwise
     pub fn is_spent(env: &Env, n: &U256) -> Result<bool, Error> {
+        Self::touch(env);
         let key = DataKey::Nullifier(n.clone());
-        Ok(env.storage().persistent().has(&key))
+        let spent = env.storage().persistent().has(&key);
+        if spent {
+            bump_entry(env, &key);
```
Fantoni0, line 671: If we already have a bump in mark_spent, I think we can remove it from her? What do you think?

contracts/pool/src/pool.rs
```diff
@@ -653,6 +688,7 @@ impl PoolContract {
     /// Returns [`Error::NotInitialized`] if the contract has no admin address
     /// stored.
     pub fn update_admin(env: Env, new_admin: Address) -> Result<(), Error> {
+        Self::touch(&env);
```
Fantoni0, line 691: Maybe put it inside soroban_utils::update_admin

contracts/asp-membership/src/lib.rs
```diff
@@ -156,6 +160,7 @@ impl ASPMembership {
     /// # Returns
     /// The Poseidon2 hash result as U256
     pub fn hash_pair(env: &Env, left: U256, right: U256) -> U256 {
+        bump_instance(env);
```
Fantoni0, line 163: I think this is a pure function and does not  read storage, so probably we can skip the bumping

Fantoni0 (changes requested):
Thanks @dkgoutham 
Left some comments. Some of them I commented only in pool.rs but also apply in Pool_gvk (and sometimes even the ASP).

contracts/asp-membership/src/lib.rs
dkgoutham: No, it still needs it. bump_instance extends the contract's own instance and code, and any call needs those live, pure or not.

contracts/pool-core/src/merkle_with_history.rs
dkgoutham: Fixed.

Fantoni0:
Thanks, @dkgoutham. The comments and new changes address most of the concerns.
However, there are 2 things that are not ready yet:

> Right now we have bump_dependency in the user path. This means a user might end up paying the rent for the pool, the verifier, and the ASPs. My suggestion would be: explore if it's possible to remove this from user calls, and have a dedicated function that anyone can call (ideally the admin) to extend the TTL. If you don't think this is possible, we can at least use extend_tt__with_limits to keep the costs bounded.
> Also applies to pool_gvk

I think this comment is unaddressed and is the most important one.
We still don't bound the price the user has to pay. 
The admin only renews the pool instance (not even the admin daatakey). I thnk it makes more sense to have the admin pay as default (and if not, users can decide to pay rent themselves to ensure livennes). But this way the admin pays first, and user pay as fallback.

> Maybe put it inside soroban_utils::update_admin
This one is also unaddressed. My point is that it will also renew the admin entry, it's already writing, and I think it's cheaper too

PR #568, opened 2026-09-08
GVK pool ASP update contract methods

v0-e:
Allow admin to update the ASP contracts.
The removed docs mention:
```rust
    // `pool` also exports `update_asp_membership`/`update_asp_non_membership`;
    // this crate deliberately does not. A `pool-gvk` pool can therefore never
    // repoint its ASP contracts after construction, unlike a `pool` one 
```

However GVK key/encryption is independent of this logic.


Fantoni0 (approved):
LGTM!

PR #569, opened 2026-09-08
Typed retention-gap error

v0-e:
Retention-gap detection matched on error message substrings. Replaced with typed errors (`IndexerError`, `Error::RetentionGap`) so callers can match on the actual cause instead of downcasting or parsing text.


Fantoni0 (approved):
LGTM!

PR #571, opened 2026-09-09
Name the signing account in the chain layer signing path

cupicmarko:
The chain layer named its signing account `user_address` and typed it `&str`.
That name belongs to the note owner, but every one of these parameters is a
payer role: the account whose sequence number is read, the envelope source, and
the address the wallet is asked to sign with. #548 split the two identities in
`types` (`NoteOwnerAddress` vs `SignerAddress`), and the signing path was still
below that split — an owner-derived string reached it unchanged.

## Change

`sign_prepared_transaction`, `auth_sign_steps` and `unsigned_tx_for_signing`
take `&SignerAddress`. `signer::local::LocalSigner` stores a `SignerAddress` and
exposes it as `signer_address()`. The wasm `WalletSigner` passes its
`SignerAddress` through instead of unwrapping it to `&str` at each call.

`tx_prepare::prepare_register` keeps `&str`: its argument is the registry key as
well as the envelope source, and which identity it should be once the two
diverge is unsettled.

### DIFF
--- a/sdk/native/README.md
+++ b/sdk/native/README.md
@@ -41,7 +41,11 @@ let client = Client::init(
 )?;
 
 let signer = Handle::from_box(
-    Box::new(LocalSigner::new("S...", "Test SDF Network ; September 2015", "G...")?)
+    Box::new(LocalSigner::new(
+        "S...",
+        "Test SDF Network ; September 2015",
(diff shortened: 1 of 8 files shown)


v0-e (approved):
lgtm!

PR #572, opened 2026-09-09
Refuse a signature the wallet made with a substituted account

cupicmarko:
When Freighter does not hold the account it was asked to sign with, it does not fail: it signs with whichever account is active and returns success, with `signerAddress` naming the substitute and nothing in the approval prompt naming the address that was requested. Both signing paths read the signature out of that reply and discarded `signerAddress`, so a signature made by the wrong account was accepted as if it were the right one.

## Change

`normalize_sign_result` returns the signature together with the address the wallet reports, and `WalletSigner::call` refuses a reply whose reported address is not the requested one. `app/js/wallet.js` applies the same check at `signTransaction`, `signAuthEntry` and `signMessage`, since the admin page reaches Freighter without going through the SDK; the rule lives in `app/js/wallet-signer-guard.js`, importless so Node can unit-test it without resolving `@stellar/freighter-api`.

A reply that carries no address is still accepted — a bare-string return has nothing to compare, and the e2e stub signers use that shape. Freighter always reports an address, so the substitution path stays covered.

The mismatch message wraps both addresses in `Sensitive` and avoids "rejected"/"denied"/"cancelled", which `normalizeWalletError` substring-matches to classify user cancellations.

### DIFF
--- a/app/js/wallet-signer-guard.js
+++ b/app/js/wallet-signer-guard.js
@@ -1,0 +1,37 @@
+/**
+ * The rule that a wallet must sign with the account it was asked to sign with.
+ *
+ * Kept in its own module, free of any import, so it is reachable from Node for
+ * unit testing: `wallet.js` pulls in `@stellar/freighter-api`, whose package
+ * exports cannot be resolved outside the browser bundle.
+ */
+
+/**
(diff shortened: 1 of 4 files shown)


v0-e (approved):
lgtm!

PR #573, opened 2026-09-09
Split prepare_register's source into owner and payer

cupicmarko:
The registry key and the transaction payer were one address. Take them as NoteOwnerAddress and SignerAddress instead: the owner becomes the Account.owner argument the registry stores and authorizes, the payer supplies the sequence number and sources the envelope.

Account::register_public_keys already held both identities separately, so it now passes each to its own role.

Replace the register test, which re-implemented prepare_register against a hand-rolled mock, with one that drives the real function over wiremock with owner != payer.

### DIFF
--- a/sdk/native/src/account.rs
+++ b/sdk/native/src/account.rs
@@ -164,10 +164,15 @@ impl<S: Storage> Account<S> {
         let fetcher = StateFetcher::new(self.rpc.clone(), self.contract_config.clone())
             .map_err(|e| Error::Other(format!("state fetcher: {e:#}")))?;
         let prepared = fetcher
-            // Registration uses the note-owner address, which is both the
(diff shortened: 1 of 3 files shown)


v0-e (approved):
lgtm!
Probably would be good to add some sort of warning to this (register) flow. The goal of splitting addresses is adding privacy / support relayers. So, IIUC, having split addresses in the register flow will ultimately link both publicly.

cupicmarko:
> lgtm! Probably would be good to add some sort of warning to this (register) flow. The goal of splitting addresses is adding privacy / support relayers. So, IIUC, having split addresses in the register flow will ultimately link both publicly.

Yes, part of a separate PR when this is exposed to the end user.

PR #574, opened 2026-09-09
Bind derived privacy keys to the note owner, not the signing account

cupicmarko:
Privacy keys are derived from a wallet signature over a constant message, so the signer owns the derived note key, encryption key and ASP membership blinding. That holds together only because `ensure_signer_is_note_owner` currently refuses any session where the two differ (#548). The moment a deployment separates the fee-paying signer from the note owner, a delegate's signature becomes the owner's key material — the delegate can decrypt the owner's notes and forge spend proofs, and the owner cannot recover without them.

This lands the binding ahead of that split, so the split can be enabled without the derivation being wrong the day it is.

## What

A key binding selected per deployment by `ContractConfig::signer_may_differ_from_owner`:

- **v1** — the existing split-free derivation, unchanged and byte-identical, retained permanently as the correct binding wherever the payer is the owner.
- **v2** — derives from a signature over `Privacy Pool Key Derivation [v2] owner=<G-address>`, verified locally against the Ed25519 key decoded from the owner's address before anything is derived from it.

Deployment-selected rather than per-account, because a client always knows which deployment it is on, but cannot tell a new account from a recovering one on empty storage. Each deployment accepts exactly the binding it derives, so recovery from the owner's secret alone always works and no account is asked to re-derive.

## Enforcement

- `schema_v3_key_binding.sql` adds `binding_version` (defaulted to 1, so existing rows are classified, not unknown) and `bound_owner`.
- Both SQL reads of `keypairs` take the required binding — the only place both platforms pass through. A wrapper-level check would cover native only, since `Storage` has two implementations and every web read goes over the worker protocol.
- An unconfigured handle **refuses**. Defaulting to v1 would accept a v1 row on a v2 deployment, which is fail-open in the one direction that matters.
- The storage worker holds the requirement, set once and refusing a conflicting change; reads and the write both take it from there, so no request can select the unverified path.
- Status moves to a metadata-only route that never touches a key blob, so "is this account onboarded" is answerable without a route that returns secrets — and the UI can distinguish *no keys* from *keys unusable here*.

## Verification

- native lib **317 passed**, web wasm32 **12 passed**, e2e-freighter unit **43 passed**
- `cargo clippy --workspace --all-targets` and `cargo fmt --check` clean
- e2e-freighter `01-connect` run 20× against a freshly provisioned profile
- migration test applies only the first two migrations, writes through the pre-v3 path, then forward-migrates — rather than trusting the DEFAULT clause
- tests pin: two owners with the same signature diverging on all three branches, v1 signatures unable to produce v2 material, non-owner signatures refused before anything is stored, clean-install recovery on both bindings, fork preserving the requirement, and an unconfigured handle refusing

## Out of scope

- **Sessions where signer ≠ owner remain refused**, natively and on web. This PR makes the derivation correct for that world; enabling it is separate work.
- **Enabling the flag on a deployment with live v1 accounts.** Those hold notes under v1 keys and re-deriving would shadow rather than replace them; making both usable needs per-note binding provenance.
- **The SEP-53 test vector is self-consistent.** It verifies the reconstruction against itself, not against a signature captured from the real Stellar CLI or Freighter, so cross-platform agreement is argued rather than pinned.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5367,6 +5367,7 @@ dependencies = [
  "anyhow",
  "rusqlite",
  "serde_json",
+ "sha2 0.11.0",
  "stellar-private-payments",
  "tracing",
  "tracing-subscriber",
--- a/app/ARCHITECTURE.md
+++ b/app/ARCHITECTURE.md
@@ -194,11 +194,29 @@ Root-level `circuits/` in the deployed site holds **legal files only** (`NOTICE.
 
 ## Keypair derivation
 
-Keys are derived deterministically from Freighter wallet signatures:
+Keys are derived deterministically from Freighter wallet signatures. Which
+construction is used is a fixed property of the deployment
+(`signer_may_differ_from_owner` in `deployments.json`, default `false`) —
+never a per-user or per-session choice:
+
+**v1 (split-free deployments, the default and permanent for them):**
 
 1. User signs `KEY_DERIVATION_MESSAGE` from `sdk/native/src/zk/encryption.rs` (`"Privacy Pool Key Derivation [v1]"`).
 2. The worker derives the BN254 note identity keypair and the X25519 encryption keypair from that signature using domain-separated hashes.
 3. Derived keys are stored in SQLite; the signature is not persisted.
+
+**v2 (deployments that separate the fee-paying signer from the note owner):**
+
+1. User signs `key_derivation_message_v2(ownerAddress)`, which names the owner address in the signed text so it is recognisable in the wallet's approval dialog.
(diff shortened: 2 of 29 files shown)


Fantoni0:
I don't think the premise of the PR description is still true after merging #575.
In general, I think this is overcomplicating things:
- We don't need, and probably would have to carefully think before modifying a new key derivation scheme. We already have 575 to separate the owner/signer figures.
- I don't think we are enforcing type safety by leveraging Rust. Most of the checks are at the string comparison level.

So, I would propose to: verify the signature against the owner's address. Then the security comes from the signature verification. We can discuss in the call today

PR #575, opened 2026-09-09
Check for the note owner where an owner signature is needed

cupicmarko:
Client::account refused any session whose signer was not the note owner, so no session could hold a divergent pair. Spending never needed the owner: pool.rs builds transact from the signer address, and the contract asks only for sender.require_auth() plus the funding transfer. The check comes off session open, and a delegated session now opens.

Two operations do need the owner itself, and each now checks for it. Key derivation's signature is the note secret, and WalletSigner asks the wallet for the single account it holds; on a divergent pair the wallet would sign with the payer and the payer's keypair would be filed under the owner's address, silently and persisted. The web client checks inside the missing-keys branch, before the wallet is asked for anything. Registration's simulated auth entry is keyed to the owner and the payer has nothing to put there, so Account::register_public_keys checks before preparing the call rather than failing at an unfillable entry.

The check itself moves from client.rs to account.rs, beside its native caller. SignerIsNotNoteOwner said a session where the two differ is not supported, which is no longer true: it now names the step that needs the signature rather than the session, and stays clear of the substrings the app's cancellation classifier matches.

A delegated session still cannot bootstrap an owner that has never derived keys — the owner opens a session once with its own account first. Lifting that means a second address on WalletSigner for message signing, which is not this change.

The web guard tests were plain #[test] in a crate whose suite runs under the wasm32 harness, so nothing ever collected them. They run as wasm_bindgen_test now, alongside a new one on each side asserting the refusal does not read as a wallet cancellation.

## Before merging

This is the change that turns delegation on, so the deposit debit wants a decision. pool.rs:448 is `token_client.transfer(&sender, &this, &amount)` and `sender` is the payer, so under a divergent pair the payer's token balance funds the deposit while the note is credited to the owner. The payer is not only covering fees, it is covering the principal. That may be the intent for a funding-agent deployment, but nothing at the call site makes the size of the grant visible.

### DIFF
--- a/sdk/native/src/account.rs
+++ b/sdk/native/src/account.rs
@@ -142,11 +142,20 @@ impl<S: Storage> Account<S> {
     }
 
     /// Register this account's public keys on the deployment-wide registry.
+    ///
+    /// # Errors
+    /// Returns [`Error::SignerIsNotNoteOwner`] on a delegated session. The
+    /// registry entry is the owner's, so simulation returns an auth entry
+    /// keyed to the owner; the payer's wallet is only asked to sign for its
+    /// own account and has nothing to put there. Checked before the call so a
+    /// delegated session stops here rather than at an unfillable auth entry.
     pub async fn register_public_keys(
         &self,
         note_public_key: Option<NotePublicKey>,
         encryption_public_key: Option<EncryptionPublicKey>,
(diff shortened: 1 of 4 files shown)


v0-e (approved):
lgtm!

PR #576, opened 2026-09-09
Add --sign-as so a separate account pays for the owner's notes

cupicmarko:
# Add `--sign-as` so a separate account pays for the owner's notes

An address plays six roles in this codebase, and the CLI fused two of them by construction: `session.rs` resolved a single `--account` alias and passed the same string to `NoteOwnerAddress::new` and `SignerAddress::new`, twice over, in `ClientSession::new` and `::new_disclosure`. #548 split the pair in `types` and #573 split `prepare_register`'s source, and 59dbe95 took the owner-must-equal-signer check off `Client::account` so a divergent session opens at all. The SDK has been able to hold two identities for three PRs; the CLI was still the place that guaranteed it never would.

## Change

`--sign-as <alias>` is a global flag beside `--account`. `--account` names the account that owns the notes; `--sign-as` names the payer that sources, signs and pays for every envelope and acts as the pool `sender`. Absent, the owner fills both roles.

`CliConfig` grows `require_signer(&owner)` alongside `require_account()`. Without the flag it returns the owner unchanged — not a second lookup that happens to agree, but the same `Account`, so the single-alias path resolves exactly what it resolved before and never shells out to `stellar keys public-key` twice. `session.rs` folds the two duplicated open blocks into one `open_account`, which is also what keeps `AliasSigner` pointed at the payer's alias rather than the owner's. Every `ClientSession` caller still passes only the owner; payer resolution is internal, so `cmd/` is untouched.

Onboarding refuses a payer outright. Its derivation signature *is* the note secret, so a session that signs with anyone else would derive the payer's keypair and file it under the owner's address. `ensure_owner_signs` runs immediately after `require_account` and before `open_storage`, so a divergent pair exits without a database, let alone a signature — the SDK's `SignerIsNotNoteOwner` fires at `register_public_keys`, which is the right place for registration and too late for local key derivation. `ensure_owner_is_signer` is a free function so the refusal is testable without the Stellar CLI, mirroring how the SDK made its own guard testable.

Two things that would otherwise rot: `validate_alias` hardcoded `--account` in both of its refusals, so it now takes the flag it is reporting on and a secret key passed to `--sign-as` is named against `--sign-as`. Withdrawal keeps defaulting its recipient to the owner — it reads `require_account()`, which is untouched — but its help said "defaults to the signing account", which after this change would read as the payer.

Both addresses in the new refusal and in the divergence log line go through `Sensitive`. The refusal carries none of the substrings the app's cancellation classifier matches on, and a test asserts that rather than leaving it to review.

## Testing

Seven unit tests: the owner-pays-for-itself default, the onboarding refusal with its redaction and its wording, self-signing accepted through both entry points, and flag-correct refusal of seed phrases and raw secret keys for each flag.

The acceptance scenario — deposit as A, `--sign-as B` withdraw to C — is not automated, because nothing in the repo drives the `spp` binary: `e2e-tests/` is a contract/SDK crate and the Makefile's e2e targets drive the web app through Freighter. It is a live-testnet check. The onboarding hole above was found and fixed against a mock `stellar` on `STELLAR_BIN`, which is the cheapest harness for anything that turns on which alias gets shelled out to.

### DIFF
--- a/cli/src/account.rs
+++ b/cli/src/account.rs
@@ -4,6 +4,10 @@
 //! named by an alias and resolved to an address through the Stellar CLI.
 //! Transaction signing is delegated to `stellar tx sign` via
 //! [`crate::signer::AliasSigner`].
+//!
+//! One of these stands behind each of the two roles a session holds: the owner
+//! that the notes belong to (`--account`) and the payer that sources and signs
+//! every envelope (`--sign-as`, defaulting to the owner).
 
 use std::path::Path;
 
@@ -11,7 +15,8 @@ use anyhow::Result;
 
 use crate::stellar_cli;
 
-/// A signing account backed by a Stellar CLI alias.
+/// An account backed by a Stellar CLI alias, in either the owner or the payer
+/// role.
 ///
 /// The alias fully identifies the account (the Stellar CLI stores each identity
 /// with its own derivation), so no HD path is threaded here.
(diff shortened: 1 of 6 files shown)


v0-e (approved):
lgtm!

PR #577, opened 2026-09-09
Account get balance

v0-e:
Adds `balance()` to `Account` to fetch non-pool balance. Seems useful enough to add to our SDK, supporting deposits and verifying withdrawals.

### DIFF
--- a/sdk/native/examples/common/mod.rs
+++ b/sdk/native/examples/common/mod.rs
@@ -32,7 +32,7 @@
 use std::path::PathBuf;
 
 use stellar_private_payments::{
-    CircuitStore, Handle, LocalProver, LocalSigner, LocalStorage, Prover, Signer,
+    CircuitStore, Error, Handle, LocalProver, LocalSigner, LocalStorage, Prover, Signer,
     blocking::{Account, Client, PrivatePool},
     chain::LocalSigner as StellarSigner,
     types::{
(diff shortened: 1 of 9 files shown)


Fantoni0 (dismissed):
Thanks @v0-e 
Mostly fine, leaving a couple of recommendations. Feel free to ignore:
- There are still a couple of CodeQL warnings, check if they are false positives.
- The merge conflict.
- There seems to be some differences on how get_token_balance, get_trustline_balance and get_account deal with non-existing accounts (some return 0, some return an error). But this might be fine, as they are different kind of assets.
- Maybe missing some tests for the new get_token/trustline_balance

v0-e:
Thank you @Fantoni0 🙏 
> There are still a couple of CodeQL warnings, check if they are false positives.

It is not prod code, so it is fine. The logs are more informative this way.

> The merge conflict.

:+1:


> There seems to be some differences on how get_token_balance, get_trustline_balance and get_account deal with non-existing accounts (some return 0, some return an error). But this might be fine, as they are different kind of assets.

It is intentional. `get_account` should actually error if the account doesn't exist. For the other ones, returning 0 is functionally correct.

> Maybe missing some tests for the new get_token/trustline_balance

Yeah, this would be way easier with #591 work, no need to mock everything.

Fantoni0 (dismissed):

v0-e:
Fixed merge conflict

Fantoni0 (approved):

PR #588, opened 2026-09-11
Isolate plan/prepare logic in examples

v0-e:
Simplifies to not use `prepare_*`, plan logic in main `deposit`, `transfer` and `withdraw` examples. That is lower level logic which is advised to not be used in most use-cases.
That API is now showcased in the new `plan` example.


Fantoni0 (dismissed):
Once the conflicts are resolved LGTM!

Fantoni0 (approved):

PR #589, opened 2026-09-11
Hold the signing account separately from the note owner in the app

cupicmarko:
The app currently stores one wallet address and relies on the SDK to use it as both the note owner and signer. This change makes those identities explicit in app session state, preparing the app to represent a separate signing account.

`App.state.wallet.signingAddress` is initialized to the connected wallet address and cleared on disconnect. A shared `accountSession()` helper supplies `userAddress` and `signerAddress` when opening accounts during wallet connection and pool selection, falling back to the owner when no signer is selected.

Onboarding remains owner-only: key derivation and public-key registration require the owner's own signature. No signer-selection UI is added, so existing user flows continue to use the same account for ownership and signing.

The helper is import-free so its behavior can be tested directly in Node without loading browser dependencies.

Validation:
- Three unit tests pass for owner fallback, a distinct signer, and disconnected state.
- Freighter E2E and all build, lint, WASM, and CodeQL checks pass.
- SDK browser E2E fails during the seed deposit with `INVOKE_HOST_FUNCTION_TRAPPED`. The same failure occurs on the parent commit on main, independently of this change.


Fantoni0:
Looks good. Thanks @cupicmarko ! :handshake: 
Just some comments:
- E2E is failing.
- The description mentions the onboarding wizard multiple times, but I don't see it in the list of modified files. Is this a leftover or some changes are missing?

cupicmarko:
>     * E2E is failing.

E2E tests are passing on newer PRs, here they seem flaky, I'll review how to address that

>     * The description mentions the onboarding wizard multiple times, but I don't see it in the list of modified files. Is this a leftover or some changes are missing?

Yes, I forgot to update the description after removing some changes. I'll update the PR description

cupicmarko:
The tests are now passing, but it is temporary, flakiness will be probably solved by #609 

Updated the PR description, no changes to the code.

v0-e (approved):
lgtm!

PR #591, opened 2026-09-11
Integration tests

v0-e:
Tests the SDK against a Stellar local network.
- new root crate since it touches several things on the repo (contracts, SDK, deploy.sh, etc);
- excluded from workspace with own CI: it needs Docker (run local network), Stellar CLI (deploy support).

Most of the changes are due to the new `Cargo.lock`.


contracts/pool-core/src/merkle_with_history.rs
```diff
@@ -7,15 +7,22 @@
...
 //! these functions.
 
 use soroban_sdk::{Env, U256, Vec, contracttype};
 use soroban_utils::{get_zeroes, poseidon2_compress};
 
-/// Number of roots kept in history for proof verification
-const ROOT_HISTORY_SIZE: u32 = 90;
+/// Ring size for the root history packed into [`TreeState`]; bigger means
+/// more stale-proof tolerance at the cost of a larger entry.
+pub const ROOT_HISTORY_SIZE: u32 = 32;
```
Fantoni0: Why did we went from 90 to 32?

integration-tests/run.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+# Run the integration-tests crate against a fresh `stellar/quickstart`
+# container: start it, run the tests, then stop and remove it — on success,
+# on failure, and on Ctrl-C.
+#
+# Usage: run.sh [test command...]   (default: cargo test --lib)
+#   run.sh cargo test --lib -- --test-threads=1
+#   run.sh cargo nextest run
```
cupicmarko: I would suggest dropping this usage line, [setting up the environment](https://github.com/NethermindEth/stellar-private-payments/pull/591/changes#diff-9a7b7d4e571672987d369edd78e332238c3e05ed69370fec2e3f9cf579f09e20R212) will race to create symlink, and with nextest it will happen. The alternative is to make symlink creation atomic, but I don't think it's really necessary.

integration-tests/src/network.rs
```diff
@@ -0,0 +1,225 @@
...
+    ///
+    /// Retries: friendbot can still 502/503 for a few seconds after
+    /// `getHealth` already reports healthy.
+    pub async fn fund(&self, address: &str) -> Result<()> {
+        const MAX_ATTEMPTS: u32 = 10;
+
+        let url = format!("http://localhost:{RPC_PORT}/friendbot?addr={address}");
+        let mut last_error = String::new();
+        for _ in 0..MAX_ATTEMPTS {
+            match reqwest::get(&url).await {
```
cupicmarko: I think that this doesn't have a timeout, so it could brick the loop. [Timeout](https://docs.rs/tokio/latest/tokio/time/struct.Timeout.html) could possibly be a better solution.

cupicmarko (dismissed):
Just two minor comments. LGTM!

integration-tests/src/network.rs
```diff
@@ -0,0 +1,225 @@
...
+fn ensure_local_vk_file(root: &Path) -> Result<()> {
+    let local_dir = root.join("deployments/local");
+    let link = local_dir.join("circuit_keys");
+    if link.is_symlink() {
+        return Ok(());
+    }
+    std::fs::create_dir_all(&local_dir).context("create deployments/local")?;
+    if link.exists() {
+        std::fs::remove_dir_all(&link).context("remove stale deployments/local/circuit_keys")?;
+    }
```
Fantoni0: Is it possible to have a race condition here? 
Different processes call these and remove the folder for the other ones?

integration-tests/src/tests/support.rs
```diff
@@ -0,0 +1,119 @@
...
+        encryption::derive_encryption_and_note_keypairs(signature.clone())?;
+    let membership_blinding = encryption::derive_membership_blinding(&signature, network)?;
+    storage.storage_mut().save_encryption_and_note_keypairs(
+        &wallet.address(),
+        &note_keypair,
+        &encryption_keypair,
+        &membership_blinding,
+    )?;
+    Ok(())
+}
```
Fantoni0: Don't we have this functionality aas part of the SDK already?

integration-tests/run.sh
v0-e: The goal is to have these tests with either test / nextest, so I fixed that symlink issue instead :+1:

integration-tests/src/network.rs
v0-e: Added :+1:

integration-tests/src/network.rs
v0-e: Also a concern of @cupicmarko, removed this from the code and moved it to setup script `run.sh`

integration-tests/src/tests/support.rs
v0-e: Indeed, that is in main now :+1:

Fantoni0 (approved):
LGTM!

cupicmarko (approved):

PR #592, opened 2026-09-11
SDK Account derive-keys

v0-e:
Adds a missing derive keys method to the SDK / account.
Makes key derivation a bit more explicit in the web side, removing it out of account creation. Even though it requires more work by the user (call derive keys), is better to make it separated since it requires storage access + signer activity (requires interaction, slow).
Also renames "user keys" to "privacy keys", since the former was a bit too generic. "Privacy" as in "privacy" pools.


Fantoni0:
Let's see if we are changing the key derivation to be SEP-53 compatible in #612 first. Will review after that :pray:

v0-e:
Key derivation is already based on that. Both CLI and Freighter prepend that prefix, only `LocalSigner` was missing the sign-message feature.

sdk/native/src/signer/local.rs
```diff
@@ -6,9 +6,13 @@ use super::Signer;
 use crate::{
     PreparedTransaction,
     error::Error,
-    types::{SignedTransaction, SignerAddress},
+    types::{KeyDerivationSignature, SignedTransaction, SignerAddress},
 };
 
+/// SEP-53 message prefix, matching the Stellar CLI and browser wallets
+/// (`stellar message sign`, Freighter's `signMessage`).
+const SEP53_PREFIX: &str = "Stellar Signed Message:\n";
```
cupicmarko: This can be reused from changes introduced in #612

sdk/native/src/blocking/account.rs
```diff
@@ -57,8 +57,12 @@ impl Account {
         block_on(self.inner.balance(asset))
     }
 
-    pub fn user_public_keys(&self) -> Result<(NotePublicKey, EncryptionPublicKey), Error> {
-        block_on(self.inner.user_public_keys())
+    pub fn privacy_keys(&self) -> Result<(NotePublicKey, EncryptionPublicKey), Error> {
```
cupicmarko, line 60: I wasn't aware of this change, I think that `Storage::privacy_keys` should be renamed to `Storage::stored_private_keys`.

sdk/web/src/workers/storage.rs
```diff
@@ -960,19 +1001,19 @@ impl Storage for StorageBridge {
...
+                StorageWorkerRequest::PrivacyKeys(user_address.to_string()),
                 1_000,
             )
             .await
         {
-            Ok(StorageWorkerResponse::UserKeys(keys)) => {
+            Ok(StorageWorkerResponse::PrivacyKeys(keys)) => {
                 let keys = keys.ok_or_else(|| {
-                    Error::Other(anyhow::anyhow!("user keys not found in worker storage"))
+                    Error::Other(anyhow::anyhow!("privacy keys not found in worker storage"))
```
cupicmarko: Could this return `Error::PrivacyKeysNotFound { user_address: user_address.to_string() }` instead? Now `client.account()` no longer derives keys, calling `account.privacyKeys()` before `derivePrivacyKeys()` reaches this branch. Web callers get "privacy keys not found in worker storage" and miss the new "derive privacy keys first”. Returning the shared error would keep native and web behavior consistent.

cupicmarko (dismissed):
Docs are after merging 612 out of sync, but otherwise, LGTM! 

Just few minor comments, and web/README.md should be updated, possibly other READMEs too.

sdk/web/src/workers/storage.rs
v0-e: Nice :+1:

sdk/native/src/blocking/account.rs
v0-e: Fixed that :+1:

sdk/native/src/signer/local.rs
v0-e: Reusing :+1:

Fantoni0 (approved):
LGTM!
Minor comment, up to you. We mix `privacy_keys` and `private_keys` in multiple files after the misc commit. I would choose 1 and stick to that

v0-e:
> We mix privacy_keys and private_keys in multiple files after the misc commit. I would choose 1 and stick to that

The idea is to use "privacy" in user-facing methods (which should also be public key material), and "private" internally  for the more secret material.

Fantoni0 (approved):

PR #593, opened 2026-09-14
Let a payer register the owner's keys, and check the payer exists

cupicmarko:
## Summary

`Account::register_public_keys` refused any session whose signer was not the note owner, but `prepare_register` beneath it already takes owner and payer as separate typed arguments, documents the divergent case, and is unit-tested with the two differing. This lifts the guard at that call site: the owner's authorization travels as the auth entry the registry's `require_auth()` asks for, while the signing account sources the envelope and pays the fee.

Because the shipped signers each fill only the auth entries for the one account they sign as, a delegated registration needs a `Signer` holding both keys. Rather than let a half-signed transaction reach the network, the envelope is checked before submission: an address-credential auth entry still empty raises `Error::MissingAuthorization` naming the address, so no fee is spent on a transaction the contract would refuse. Teaching `LocalSigner` and the web SDK's wallet signer to fill the owner's entry is deliberately left out — it needs either a new `Signer` trait method or a second signer handle on the session, and in the browser it collides with the still-open account-switch policy.

The derivation guard is untouched in substance — `ensure_signer_is_note_owner` must keep protecting key derivation, where the signature *is* the note secret. Removing its only in-crate call site left it dead, so it becomes public API instead: the derivation paths that need it live in the web SDK and the CLI, and each restates the rule today.

Separately, the paying account is now checked before envelope building. A Stellar account exists only once it is funded, and envelope building reads the payer's sequence number, so an absent account used to surface late as whatever the RPC returned for a missing ledger entry. `Error::PayingAccountNotFound` names it instead, with the address wrapped in `Sensitive` like every other address in a user-facing error here, and worded clear of `rejected` / `denied` / `cancelled` so the web app's substring classifier does not read a precondition failure as a wallet cancellation.


Fantoni0:
So anyone can relate accounts A and B together?

PR #601, opened 2026-09-14
Remove args from `register_public_keys`

v0-e:
Simplifies `Account::register_public_keys()`, removing the pub key args. There isn't really a scenario where the user isn't going to register another's public keys. *Tx signing delegation (relayers) is a different topic and handled through another mechanism.


Fantoni0 (dismissed):
Nice 💯 @v0-e

Fantoni0 (approved):

PR #602, opened 2026-09-15
Fix the UB detection workflow build

### LINKED ISSUE #352: Fix UB action
https://github.com/NethermindEth/stellar-private-payments/actions/workflows/ub-detection.yml

zexoverz:
The nightly job never got as far as running Miri, it failed to compile. `construct_uint!` from uint 0.10.1 expands to `std::isize::MAX` and `<int>::max_value()`, which nightly now marks deprecated, and under `-Dwarnings` that is a hard error reported at the invocation site in `amounts.rs`. There is no newer uint release and their master has no fix either, so the `allow` already on that module gains `deprecated` instead of the lint being turned off crate wide. The other break is in `e2e-tests`, where `all_transact_graphs_prove_and_verify` carried both `#[ignore]` and `#[cfg_attr(miri, ignore)]` and nightly reports the second as unused, so it is dropped.

Checked with the workflow's own command on nightly (`RUSTFLAGS=-Dwarnings`, `MIRIFLAGS=-Zmiri-strict-provenance`, `cargo +nightly miri test --lib` for the app-core and contracts-light groups). Both compile now and contracts-light ran clean as far as I let it go. One thing I could not settle from a Mac: `blocking::runtime::tests::block_on_sync_context` builds a tokio runtime and Miri here refuses `kqueue`. Linux Miri has epoll so it may pass on the runner, and if the scheduled run trips on it the follow-up is `#[cfg_attr(miri, ignore)]` on that one test.

Fixes #352


Fantoni0 (approved):
LGTM!

Issue #376, opened 2026-07-09
Move immutable config from persistent to instance storage to save gas


Fantoni0:
consider any storage optimizations which can be introduced to optimize contracts resources usage (e.g. may be some data can be put to another storage or embedded at compile time etc)

PR #607, opened 2026-09-16
measure the footprint and host cost of every entry point

dkgoutham:
## What changed and why

The contracts have never had a test that reports what a call costs on the ledger. Fees on Stellar are driven by the footprint, the set of ledger entries a call reads and writes, and by the bytes written and the rent those bytes accrue, so a storage layout that touches thirty entries per call is thirty times the read fee of one that touches one. Nothing in the repository measured that, and the layout grew by adding a persistent entry whenever something needed storing.

This adds a cost table in `e2e-tests`. Each row invokes one entry point once, against a fresh deployment in the test host, and records what the host metered for that invocation: the ledger entries the call touched, the entries and bytes it wrote, the host instructions and memory it used, and the rent it accrued on persistent entries. The ledger runs with mainnet's entry lifetimes, so the rent column reflects the lifetimes a real call is charged for. The rows cover a pool deposit, transfer, and withdrawal against a fresh tree, a transfer against a root one transaction old, a transfer under both policies, `get_root`, a GVK transfer, a membership insert, a non-membership insert and delete, and a public key registration, plus one transfer with a real Groth16 proof over a 64-leaf tree, which is the only row whose instruction count includes the pairing check.

Every row pins its entry and write counts, so a change to the storage layout shows up as a diff of the pinned numbers rather than as a surprise on testnet. The baseline the table records on `main`: a transfer touches 55 entries and writes 25, a deposit 59 and 27, a membership insert 27 and 13, a non-membership insert 14 and 10. The four pull requests stacked on this one bring those down, and each one's diff of the pins says by how much.

The harness proves a transaction the same way the 2-in-2-out end-to-end test does, so the first commit lifts that test's proving code into `prove_transaction`, `ProvenTransaction::into_proof`, and `transact` in the shared test utilities, and the end-to-end test calls them instead of carrying its own copy.

Two cost terms are out of reach here, and the module doc says so: the contracts run as native test contracts, so their own arithmetic is not metered, and transaction size, with the bandwidth and history fees it drives, is not modeled by the host.

Part of #376.


e2e-tests/src/tests/costs.rs
```diff
@@ -0,0 +1,522 @@
...
+
+fn assert_pinned(rows: &[Row]) {
+    assert_eq!(
+        rows.len(),
+        EXPECTED.len(),
+        "every pinned row must be measured"
+    );
+    let mut mismatches = std::vec::Vec::new();
+    for row in rows {
+        match EXPECTED.iter().find(|(path, ..)| *path == row.path) {
```
v0-e, line 412: Just a small thing: perhaps is better to validate if every entry of EXPECTED was matched, rather than the other way around. Maybe a small uniqueness check on all EXPECTED paths before the loop?

I could also propose moving the expected vs updated value checks to each separate fn, and return err if mismatch, collect mismatch and log that. Perhaps better than this displaced string-based matching flow.

v0-e (commented):
looks good! just a nit

v0-e (approved):

e2e-tests/src/tests/costs.rs
```diff
@@ -0,0 +1,522 @@
...
+        let fee = r.estimate_fees(
+            &FEES,
+            FEE_PER_RENT_1KB,
+            PERSISTENT_RENT_DENOMINATOR,
+            TEMPORARY_RENT_DENOMINATOR,
+        );
+        format!(
+            "| {} | {} | {} | {} | {} | {} | {} | {} | {} | {} |",
+            self.path,
+            r.memory_read_entries,
```
cupicmarko, line 132: I don't this `Entries` column, it's `memory_read_entries` and what I see from the documentation, it doesn't affect fees because it's read from memory. I compared #608 and I think that reduction is really small, but I might be wrong.

e2e-tests/src/tests/costs.rs
```diff
@@ -0,0 +1,522 @@
...
+    ("pool get_root", 3, 0),
+    ("pool-gvk transact, transfer, view-only", 57, 25),
+    ("asp-membership insert_leaf, first leaf", 27, 13),
+    ("asp-non-membership insert_leaf, ninth key", 14, 10),
+    ("asp-non-membership delete_leaf, one of nine", 14, 7),
+    ("public-key-registry register, first registration", 4, 2),
+];
+
+fn assert_pinned(rows: &[Row]) {
+    assert_eq!(
```
cupicmarko, line 405: This could pass with duplicate and missing rows, not completely sure if it's caught after this assert

e2e-tests/src/tests/utils.rs
```diff
@@ -569,6 +570,178 @@ pub fn generate_proof(
...
+    let keys = case
+        .inputs
+        .iter()
+        .map(|input| NonMembership {
+            key_non_inclusion: scalar_to_bigint(derive_public_key(input.priv_key)),
+        })
+        .collect::<Vec<_>>();
+
+    let witness = prepare_transaction_witness(&case, leaves.clone(), LEVELS)?;
+    let public_amount = Scalar::from(ext_amount);
```
cupicmarko, line 722: Beware of `Scaller::from`, it wraps negative values. Probably need tests for `prove_transaction`

cupicmarko (approved):
Left few comments, overall good!

Beware that CI is not run on other PRs in the stack

Fantoni0 (approved):
LGTM! Good job @dkgoutham

Fantoni0 (approved):

Fantoni0 (approved):
LGTM!

e2e-tests/src/tests/costs.rs
cupicmarko: I think that this comment was addressed in [5252160](https://github.com/NethermindEth/stellar-private-payments/pull/607/commits/52521602224ff5c1a95ad65dfca0e994cdb90fbc)

PR #608, opened 2026-09-16
derive zero hashes instead of storing them

dkgoutham:
## What changed and why

A Merkle tree of depth `n` has a fixed "zero hash" per level: the hash of an empty leaf at level zero, and at each level above, the hash of the level below paired with itself. Every pool and every membership set computed those at construction and wrote one persistent `Zeroes(i)` entry per level plus one for the root, 21 for the 20-level pools and 11 for the 10-level sets the deploy script creates, then read one of them back on each insertion to fill the empty sibling. The values never change: level zero is the Poseidon2 hash of `"XLM"` and each level above is the compression of the level below with itself, the same on every deployment and on every network.

This puts that table in `soroban-utils` as a constant, `ZERO_HASHES`, 33 big-endian field elements, enough for the 32-level maximum the constructors accept, behind `zero_hash(env, level)`, which returns `None` at or above the table's size. The pool core and the membership contract read it instead of the ledger, so the `Zeroes` key is gone from both key enums and the constructor writes only the filled subtrees and the first root. Reading a constant from the Wasm costs no ledger entry and no rent, while a persistent entry costs both on every call that touches it and can expire.

Per the cost table on the first pull request in this stack, a pool transfer drops from 55 entries to 36 and a membership insert from 27 to 17. The write counts do not change here; the writes go in the next two pull requests. `get_zeroes`, which returns the whole table as a vector, stays because the circuit crate documents against it and nothing else in the workspace is served by removing it.

## Breaking changes

The storage layout changes. The contracts have no upgrade entry point and nothing migrates an existing entry, so a pool or set deployed from `main` is replaced, not upgraded. The SDK cannot read a contract built from this pull request until the last pull request in the stack, which is why the stack is reviewed and merged together.

Part of #376.


v0-e (approved):
lgtm!

cupicmarko (approved):
LGTM!

Fantoni0 (approved):
LGTM

PR #609, opened 2026-09-16
derive the root ring slot and pack the filled subtrees

dkgoutham:
## What changed and why

Two parts of the pool's Merkle tree cost far more ledger entries than the information they hold.

The root history is a ring of 90 slots, one persistent `Root(i)` entry each, so a proof made against a root a few transactions old still verifies. `is_known_root` scanned that ring from slot zero upwards until it found the root, so a proof against the previous root, the common case for a wallet that proved while another transaction landed, read most of the ring: 124 entries for a transfer one transaction old. The walk now starts at the newest slot and steps backwards, wrapping from slot zero to slot 89, which finds a recent root in one or two reads; once the last commit packs the ring into one entry, the walk over what is by then an in-memory vector becomes a single `contains`. The newest slot itself was a stored `CurrentRootIndex` entry, written on every insertion; it is now derived from the leaf count, since the tree inserts two leaves per transaction and the slot is the transaction count modulo 90, and the entry is gone.

The filled subtrees, the left-sibling hashes along the insertion path that the next insertion pairs with, were one persistent `FilledSubtrees(i)` entry per level. Each insertion read every level and rewrote several. They are now one `Vec<U256>` under a single key, read once and written once. The membership contract packs its own subtrees the same way.

The last commit folds the leaf counter and the root ring into that same entry, `State`. A Soroban transaction declares the ledger keys it will touch when it is simulated, and the host fails it at apply time if it reaches outside that declaration. The ring's per-slot keys moved with the leaf count: two wallets that simulated against the same ledger both declared `Root(slot)`, the first to land advanced the counter, and the second wrote `Root(slot + 1)` and failed with a footprint error although nothing in it was wrong. The tree is now one entry whose key and size never change, so both land. `init` fills all 90 ring slots with the empty root so the entry's size is fixed from the first ledger, and every insertion writes the entry back whether or not the path changed a sibling, because a write that depends on the tree's state is the same hazard in a smaller form. The pool tests assert the shape directly: the persistent key set is the same after `init` and after insertions, and the entry's XDR length is the same after `init`, after one insertion, and after 91.

Per the cost table, a transfer goes from 36 entries and 25 writes to 14 and 4, a deposit from 40 and 27 to 18 and 6, and the stale-root transfer from 124 entries to 14, the same as a fresh one; the harness asserts that equality. The bytes a transfer writes rise from 3236 to 4444, because the 90-slot ring is rewritten on every insertion, but the 21 write entries it no longer pays for cost more at 2,500 stroops each than the extra 1,208 bytes do at 875 stroops per kilobyte. A membership insert goes from 17 entries and 13 writes to 8 and 4, and its written bytes from 1648 to 752. `pool.wasm` after `stellar contract build --optimize` is 29,982 bytes against 30,328 on `main`.

The pool tests cover the ring: the slot follows the leaf count, it wraps after ninety inserts and the roots in slots 1 and 89 are still found through the wrap, the previous root is known after one insert, and the empty root stops being known at the ninetieth insertion, prefilled slots or not. The membership contract's write skip is asserted through the host's write count: an insertion whose path is all right children writes one entry fewer than one with a left child.

## Breaking changes

The storage layout changes. The pool's tree is one persistent `State` entry holding the leaf counter, the filled subtrees, and the root ring; `CurrentRootIndex`, `NextIndex`, the per-level `FilledSubtrees(i)` keys, and the per-slot `Root(i)` keys are gone. The membership contract's per-level `FilledSubtrees(i)` keys become one `FilledSubtrees` entry. The contracts have no upgrade entry point and nothing migrates an existing entry, so a pool or set deployed from `main` is replaced, not upgraded.

Part of #376.


contracts/asp-membership/src/lib.rs
```diff
@@ -190,24 +192,32 @@ impl ASPMembership {
         }
         let mut current_hash = leaf.clone();
 
+        let mut filled: Vec<U256> = store
+            .get(&DataKey::FilledSubtrees)
+            .ok_or(Error::NotInitialized)?;
+        let mut filled_changed = false;
```
Fantoni0, line 198: Do we need filled_changed? I think it only stays false if is_right is true for all levels (full tree, what we already checked above)
Can we let the unconditional write we had before?

contracts/pool-core/src/merkle_with_history.rs
```diff
@@ -28,20 +38,41 @@ pub enum Error {
...
+    pub roots: Vec<U256>,
+}
+
 /// Storage keys for Merkle tree persistent data
 #[contracttype]
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum MerkleDataKey {
     /// Number of levels in the Merkle tree
     Levels,
-    /// Current position in the root history ring buffer
```
cupicmarko, line 37: `CurrentRootIndex`, `NextIndex` and `Root` were removed, but `StateFetcher::contracts_data` still reads them, fetching state will fail for a newly deployed pool.

cupicmarko (changes requested):

contracts/pool-core/src/merkle_with_history.rs
cupicmarko: Resolved by 611

cupicmarko (approved):

contracts/asp-membership/src/lib.rs
Fantoni0: I think this still applies
Edit: 
I thought it would maintain the thread.
It was my previous comment mentioning that I don't think we need the filled variable, and we can have an unconditional write as before

Fantoni0 (changes requested):

contracts/asp-membership/src/lib.rs
cupicmarko: It was updated in the code.

Fantoni0 (dismissed):
LGTM!

v0-e:
Since we are redeploying here, lets redeploy to have https://github.com/NethermindEth/stellar-private-payments/pull/609/commits/25130d1f8120a32ccfe110a5c4215162f9bab43d changes?

cupicmarko:
@v0-e We should redeploy again?

v0-e (approved):

cupicmarko (approved):

v0-e (approved):

Issue #372, opened 2026-07-09
Proper governance of the contracts (separation of roles, multisig 3-5, pausable operations, timelock)


Fantoni0:
also take into consideration storage TTL prolongation - how to properly administer that?

Fantoni0:
In Stellar, [storage](https://developers.stellar.org/docs/learn/fundamentals/contract-development/storage/persisting-data) needs to be paid for regularly, as a form of rent. We mostly use Persistent storage, as we need our data (pool configuration, merkle tree, and nullifiers) to always be available. However, even persistent storage has a time-to-live (TTL) that must be extended (paid for). If not, the state gets archived, and the next user who reads it has to pay the unarchival fees.

Right now, we are not extending any TTL, which means that all our data will be archived. This creates unfair lottery fees for users, as some users might interact with the pool with significantly higher fees than others. As per our conversations with SDF, the admin should be responsible for extending the TTL (let's see if we need to add dedicated methods for that). But users must be able to extend it if needed (in case the admin forgets) to ensure pool liveness.

PR #610, opened 2026-09-16
keep contract configuration in the instance entry

dkgoutham:
## What changed and why

Every contract call loads the contract's instance entry before it runs, so a setting kept in that entry's storage map costs no extra ledger entry in the footprint and shares the instance's lifetime instead of having its own. The values the constructor writes once and every call reads were persistent entries, one footprint slot each and one rent line each: the tree depth in the pool core, the pool's token, verifier, association set addresses, deposit cap, and policy flags, the GVK pool's admin view key and mode on top of those, and the membership and non-membership sets' depth and current root. All of them move into the instance entry. The two GVK setters that `main` gained in #568, `update_asp_membership` and `update_asp_non_membership`, write the instance too, and so do the pool's, and the GVK pool's `get_gvk_mode` reads it from there.

The trade is the instance entry's size. Every call now writes the whole instance entry when it changes any part of it, and the sets' root changes on every insertion, so the bytes a membership insert writes rise from 752 to 840 and the non-membership rows by 52, against two entries fewer in the footprint for a membership insert, one for a non-membership insert, and six for a pool transfer. Per the cost table, a transfer goes from 17 entries to 11 and a deposit from 21 to 14, a GVK transfer from 19 to 11, a membership insert from 8 to 6, and a non-membership insert from 14 to 13.

`Admin` stays a persistent entry in all four contracts. The governance rewrite tracked in #372 replaces the admin accessors, and moving the key under it here would make that branch resolve the same lines twice; the one extra entry per call goes when that work lands.

Each contract's tests read the moved keys back through `env.storage().instance()`, and the setters' tests check the instance holds the new address.

## Breaking changes

The storage layout changes: the keys above are instance keys, and a client that reads them as persistent contract data finds nothing. The contracts have no upgrade entry point and nothing migrates an existing entry, so a pool or set deployed from `main` is replaced, not upgraded. The SDK reads the new keys from the last pull request in this stack.

Part of #376.


cupicmarko (approved):
I see that it depends on #610, on its own, this PR fails with `MissingRequiredContractKeys`, so 610 and 611 should be merged together.

Fantoni0 (approved):
Additionally,  some parts of the SDK need to be updated (e.g: rpc and conrtact_state). THey still read the moved configuration as Persistent data. 
It's fine as this does not target main, but needs to be addressed before main.

PR #611, opened 2026-09-16
read contract configuration from the instance entry

dkgoutham:
## What changed and why

The SDK read every contract setting as its own persistent ledger entry. After the previous pull request in this stack those settings live in the storage map of the contract's instance entry, which `get_contract_data_bulk` already fetched for every contract and then ignored. The bulk read now flattens that map into the contract's result map, keyed by the symbol each `DataKey` variant is stored under, so the state reader finds `Token`, `PolicyFlags`, `Levels`, the sets' `Root`, and the rest under the names it already used, and a contract whose instance lacks one of them reports it missing by that name as before.

Keys that only some deployments have, `AdminViewKey` and `GvkMode` on the GVK pool, used to need the request's `optional_enum_keys` list so that a plain pool did not fail the bulk read; they now arrive with the instance or not at all, so that list and its plumbing are gone. The pool's leaf counter and root ring arrive together in the `State` entry, so the reader decodes that entry, derives the ring slot from the counter and the ring's length the way the contract does, and takes the root from the slot. The second round trip that fetched `Root` at a slot computed from an earlier read, and the retry loop that re-read `NextIndex` to detect drift between the two, are gone, along with the `valued_keys` request field that only the ring used. `PoolInfo::merkle_current_root_index` keeps its value, so the web SDK getter and the CLI overview do not change.

The Freighter end-to-end test read the deposit cap through the persistent `MaximumDepositAmount` key, which no longer exists. It now reads the instance entry with `stellar contract read --key-xdr AAAAFA==` and picks the cap out of the storage map.

The SDK tests cover the flattening (one-symbol vector keys are read, other key shapes and a missing map yield nothing), the ring slot including an empty ring, the decoding of the packed tree entry, and a mocked RPC response whose instance settings arrive under their own names.

## Breaking changes

The SDK reads the layout from the previous pull request and no longer reads a pool or set deployed from `main`. Between the second pull request in the stack and this one the SDK cannot read the contracts on the branch, and no test deploys them, which is why the stack is reviewed and merged together.

Closes #376.


sdk/native/src/chain/contract_state.rs
cupicmarko, line 1: We need a test that will use actual entries in `contracts_data_for_pool` that checks the complete read path, using actual storage.

cupicmarko (approved):
Approval for all the PRs in stack, I'll go back to #609 to change my "requested changes" to "approved" because I think that all the issues I found previously were resolved by this last PR.

Fantoni0 (approved):
LGTM!

PR #612, opened 2026-09-16
Verify the key-derivation signature against the note owner

cupicmarko:
The key-derivation signature is the note secret, so keys derived from any other account's signature get filed under the note owner and are silently wrong. Today that is guarded by comparing addresses, which cannot rule it out: a wallet can sign with a different account than the one it was asked for, and its reply need not say which account it used.

This checks the signature itself. Before anything is derived or stored, the signature is verified against the owner's own Ed25519 key over the SEP-53 payload (`SHA-256("Stellar Signed Message:\n" || message)`), in both places keys are derived: CLI onboarding and the web SDK's account session. The address comparison stays as a cheap early exit that spares a pointless wallet prompt; the signature check is what binds the keys to the owner. Verification is strict, as the network's own is: a small-order owner key or `R` is refused, where the lenient check accepts a forgery anyone can produce for such a key.

Both callers now refuse before they store. In the CLI, the verify-derive-store step is `save_owner_keys`, separate from fetching the signature through the Stellar CLI, and its tests assert that another account's signature leaves the wallet database untouched. In the browser, `e2e_foreign_derivation_signature_is_refused` opens a session for account A with a wallet that signs the derivation message as B, twice: both attempts must fail with the owner error, so the first refusal cannot have stored keys for A to find. The unit tests cover the signature itself: a genuine `stellar message sign` vector, another account's signature, another message, a forged signature for a small-order key, a wrong-length signature, and an owner that is not an account address.

This replaces #574 with only the part of it that was needed. The derivation message and scheme are unchanged — no v2 message, no per-deployment flag, no storage migration — so existing keys are unchanged too, and no contract is touched.

The browser e2e stub signer returned a fixed dummy blob from `signMessage`, which would no longer verify; it now signs the SEP-53 payload with the test account's own key, as a real wallet does. That also ends a long-running source of e2e flakiness. Because the blob was the same for every account in every run, all of them derived one note keypair, so each run's balance included every other run's seed deposits, and an overlapping run landing a deposit mid-test broke the exact balance assertions. Keys now follow the ephemeral accounts, one set per run. While here, `SignerMode::Signing` drops its account so a test signer always signs as the account its session was opened for, rather than allowing a pair that would derive as one account and sign transactions as another.


sdk/native/src/zk/encryption.rs
```diff
@@ -50,6 +51,43 @@ use x25519_dalek::{PublicKey, StaticSecret};
...
+        .try_into()
+        .map_err(|_| anyhow!("key-derivation signature must be 64 bytes"))?;
+    let owner = stellar_strkey::ed25519::PublicKey::from_string(owner_address)
+        .map_err(|_| anyhow!("note owner is not a valid account address"))?;
+    let owner_key = VerifyingKey::from_bytes(&owner.0)
+        .map_err(|e| anyhow!("note owner key is not a valid Ed25519 key: {e}"))?;
+
+    let digest: [u8; 32] = Sha256::digest(sep53_payload(message)).into();
+    owner_key
+        .verify(&digest, &DalekSignature::from_bytes(signature))
```
Fantoni0: Let's use verify_strict here. Just to be safer

sdk/web/src/client/e2e_tests.rs
```diff
@@ -216,21 +208,45 @@ enum SignerMode {
...
+            "E2E_ACCOUNT_{}_SECRET must be a valid S… key",
+            account.label
+        )
+    })
 }
 
 /// Build a test signer object with all three methods `WalletSigner` requires.
-fn signer_with_mode(mode: SignerMode) -> JsValue {
+fn signer_with_mode(account: TestAccount, mode: SignerMode) -> JsValue {
     let signer = Object::new();
```
Fantoni0, line 246: I know it's not a problem right now. But Would it be possible to make a bad ussage of this by passing TestAccountA, Mode:signing(TestAccountB))?

sdk/native/src/zk/encryption.rs
```diff
@@ -407,6 +445,88 @@ fn decrypt_note_data(private_key_bytes: &[u8], encrypted_data: &[u8]) -> Result<
     }
 }
 
+#[cfg(test)]
```
Fantoni0, line 450: Do we have tests that ensure rejection of missmatched signatures?

sdk/native/src/zk/encryption.rs
```diff
@@ -50,6 +51,43 @@ use x25519_dalek::{PublicKey, StaticSecret};
 /// Message signed to derive both privacy keypairs.
 pub const KEY_DERIVATION_MESSAGE: &str = "Privacy Pool Key Derivation [v1]";
 
+/// Prefix a SEP-53 wallet puts in front of a message before hashing it.
+const SEP53_MESSAGE_PREFIX: &str = "Stellar Signed Message:\n";
+
```
Fantoni0, line 56: Does this breaks backwards compat of the Key dereivation and requires a re-deployment

Fantoni0 (changes requested):
Looks good and solves a real problem. Thanks @cupicmarko 
Left some questions!

Fantoni0:
Also, sorry, forgot to mention: docs and code comments should reflect the changes in key derivation. I saw some old references

sdk/native/src/zk/encryption.rs
cupicmarko: Done.

sdk/web/src/client/e2e_tests.rs
cupicmarko: Nice catch. `SignerMode::Signing` no longer carries an account, so the signer is always built for the one account `open_account_with` takes. The message, transactions and auth entries all sign as that account.

sdk/native/src/zk/encryption.rs
cupicmarko: Tests were already in `owner_signature_tests`, I added few more:
* `another_accounts_signature_stores_nothing` checks both the refusal and that no keys were stored
* `e2e_foreign_derivation_signature_is_refused` opens account A twice, with a wallet that signs as B

sdk/native/src/zk/encryption.rs
cupicmarko: I don't think it does, keys already stored are never re-checked, the check runs only when deriving. I think that only signing with wallet that isn't SEP-53 is going to fail.

Fantoni0 (dismissed):
Some non-blockling thigns you might want to check:

- The sdk/web/tests/README.md still contains references to the old signMesage approach.
- CLI calls save_owner_keys wichh verifies, derives and stores at once.  The web app does not. Would it make sense to have verify_owner_signature inside derive_save_user_keys to it mirrors what happens in the CLI?

v0-e (approved):
lgtm!

PR #613, opened 2026-09-16
Feature/app signing account picker

cupicmarko:
## Summary

Separate the account that owns private notes from the account that signs and pays for transactions.

The connected note owner remains fixed while transfers and withdrawals can use another Freighter account as the transaction source. Deposits continue to be signed and funded exclusively by the note owner.

## What changed

- Track the note owner and signing account independently throughout the app and SDK session.
- Add “Sign and pay with” selectors to Move Funds and Advanced transactions.
- Require an explicit signer for non-deposit operations.
- Always use the note owner for deposits and net public deposits.
- Reopen cached pool sessions when the selected signer changes.
- Offer Freighter’s active account as a signer without automatically changing the note owner.
- Remember signing accounts per owner in browser storage and allow removing them under Settings.
- Add an explicit flow for changing the note owner.
- Add recipient selectors using the owner, saved accounts, or a manually entered address.
- Show both the note owner and signer in transaction confirmations when they differ.
- Warn when signing or withdrawing with the deposit account could link private activity on-chain.
- Validate pasted and unfunded signing accounts before proving.
- Improve errors when the selected signing account is unavailable in Freighter.
- Hide the signing-account card for deposits and make the initial picker prompt non-selectable.

## Session behavior

The note owner is remembered across page loads so switching Freighter’s active account for signing does not silently change ownership. Saved signing accounts are scoped to that owner and persist locally until removed from Settings or the owner is changed.

## Test coverage

- Added unit coverage for owner/signer separation, persistence, account selection, privacy warnings, and malformed stored data.
- Added transaction-envelope parsing to verify the actual source account on-chain.
- Added a Freighter E2E scenario where account C owns the notes, account C funds the deposit, and account D signs and pays for the withdrawal.
- Expanded failure-mode coverage for invalid, missing, unfunded, and unavailable signing accounts.
- Updated Freighter profile provisioning to import both required test accounts and verify the expected owner/signing setup.


app/js/signing-account.js
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * A withdrawal names its recipient on-chain and is sent by the signing account,
+ * so paying the owner from another account puts both addresses in one
+ * transaction.
+ *
+ * @param {{ owner: string | null, signer: string | null, recipient: string | null }} withdrawal
+ * @returns {boolean}
+ */
+export function withdrawalLinksAccounts({ owner, signer, recipient }) {
+    return Boolean(owner && signer && signer !== owner && recipient === owner);
```
Fantoni0: This is a valid case that mightl ink the signer and the owner/recipient. But the case we also need to check is signer =owner=recipient.

Fantoni0: ALso I think the prefilled iaccount in withdraw is always the owner. Which is fine, but will probably always triggere this

app/js/signing-account.js
```diff
@@ -0,0 +1,118 @@
...
+ *   owner: string | null,
+ *   signers?: string[],
+ *   active?: string | null,
+ * }} choice
+ * @returns {string | null}
+ */
+export function chosenSigner({ selected, owner, signers = [], active = null }) {
+    if (!selected) return owner;
+    if (signers.includes(selected)) return selected;
+    return selected === activeSuggestion({ active, owner, signers }) ? selected : owner;
```
Fantoni0: Similarly, here I think the fallbak is the owner. Might be because of how Freighter works, but the default is not too privacy-firendly

Fantoni0 (changes requested):
Looks pretty good @cupicmarko !
Let's not allow depositing from a different address for now. Deposits MUST come from the same account that signs. It could be good for privacy, and a nice feature, but it can mess many of the compliance tools we have. 
I'll upload somme images about UI comments later.

Fantoni0:
Deposit mode should not support the sign&pay modal
image

Fantoni0:
Maybe adding some hints here on why it's important to use separate accounts.
Can be tackled in future PR
image

Fantoni0:
If I withdraw with the same account I deposited (worst case for privacy). I get no warning:
image

Fantoni0:
I think this warning should be more explicit about the main problem: You are withdrawing to the same account that deposited the funds. 
The relation between the owner/signer is bad, but it's worse that we are withdrawing to the same account that made the deposit.
Maybe we can add a short note on how to do better: withdraw to a non-related account.
image

Fantoni0:
For future Prs would be nice to be able to add short descriptions to the added accounts. E.g: signer, recipient, gazillionXLMHolder etc.

image

app/js/signing-account.js
cupicmarko: Changed pre-selected account, also there are warnings now on any of the owner signing a transaction or being a recipient

app/js/signing-account.js
cupicmarko: Changed default

cupicmarko:
@Fantoni0 Done everything except for labels, they're a bit bigger decision to make, how are they managed, where they should be shown etc. Added for now a default label for the owner only.

Deposits don't allow for different signer, connection between account leaks are now indicated when owner account is selected and show explanation.

Fantoni0:
image
Nitpick: In advanced mode, when I make a pure deposit, the div is still visible

Fantoni0:
image
Nitpick: Is it possible to have the "Choose an account" not appear as an option? I think it's confusing with "Enter another address".
Also the list seems to be persistent across multiple access to the page. Is it intentional?

Fantoni0 (dismissed):
Great @cupicmarko !
LGTM!
I left 2 small nitpick comments, feel free to create good-first-issues for them. I think it's ready to be merged

cupicmarko:
Fixed the superfluous div in advanced view and updated the list of accounts. The list is intentionally fixed, owner and every account used for signing are stored in local storage so we can actually not rely on what is selected in freighter, every signing leaves that account selected.

Fantoni0 (approved):

PR #617, opened 2026-09-17
Admin UI GVK account graph

v0-e:
Adds a new view, focusing on PKs and transactions among them.
Some minor polishes also.


app/js/admin-gvk.js
```diff
@@ -797,6 +827,230 @@ function collectNotes(rows) {
...
+      for (const slot of inputs) {
+        const pk = normalizeFieldKey(slot.audited.note.pk);
+        if (!pk) continue;
+        const node = ensureNode(pk);
+        node.withdrawn += parseFieldAmount(slot.audited.note.amount);
+        node.withdrawCount += 1;
+        node.txIndices.add(index);
+      }
+      continue;
+    }
```
Fantoni0, line 901: We can have partial withdrawals. Where a part is withdrawn, and the other part remains as a new note in the pool. 
Here we only check inputs but we are not checking if any of the outputs belongs to the same pk

Fantoni0 (changes requested):
Looks really good @v0-e 
I know this one is difficult but the default view has notes appear cluttered 

Image

would it be possible to adapt the x-axis to avoid it?

v0-e:
> I know this one is difficult but the default view has notes appear cluttered

Not sure if there is anything to do. I can explore some sort of blurring method and do that over clusters, but unsure on the benefits.
I just added now the missing zoom to this view that the note graph also has. That should help on the un-clustering at least. :+1:

> would it be possible to adapt the x-axis to avoid it?

Best to show the full activity by default.
Maybe some sort of breaks/skips on empty activity. But it will not improve much on longer deployments.

app/js/admin-gvk.js
v0-e: Fixed that :+1:

app/js/admin-gvk.js
```diff
@@ -1202,6 +1525,562 @@ function renderGraph(rows, containerWidth, filters) {
...
+  if (dot.deposited > 0n) {
+    row.appendChild(el('div', 'text-[10px] uppercase tracking-wide text-emerald-400', 'This deposit'));
+    row.appendChild(el('div', 'font-mono text-slate-200', formatTokenAmount(dot.deposited)));
+  } else if (dot.withdrawn > 0n) {
+    row.appendChild(el('div', 'text-[10px] uppercase tracking-wide text-rose-400', 'This withdrawal'));
+    row.appendChild(el('div', 'font-mono text-slate-200', formatTokenAmount(dot.withdrawn)));
+  } else {
+    row.appendChild(el('div', 'text-[10px] uppercase tracking-wide text-cyan-400', 'This transfer'));
+    if (dot.transferInCount > 0) row.appendChild(el('div', 'font-mono text-slate-200', `Received ${formatTokenAmount(dot.transferIn)}`));
+    if (dot.transferOutCount > 0) row.appendChild(el('div', 'font-mono text-slate-200', `Sent ${formatTokenAmount(dot.transferOut)}`));
```
Fantoni0: IF dot comes from buildPkTimelineEvents I don't tink we have these fields. Is it possible to happen?

app/js/admin-gvk.js
```diff
@@ -797,6 +827,252 @@ function collectNotes(rows) {
...
+    }
+
+    if (effectiveKind === 'withdraw') {
+      for (const [pk, amount] of netWithdrawnByPk(inputs, outputs)) {
+        const node = ensureNode(pk);
+        node.withdrawn += amount;
+        node.withdrawCount += 1;
+        node.txIndices.add(index);
+      }
+      continue;
```
Fantoni0: Here, if we have a withdraw where some outputs are under a different pk (transfer) I don't think they are represneted properly as netWithdranByPK only checks against its own same pk outputs

Fantoni0 (changes requested):
Thanks @v0-e , just some comments

app/js/admin-gvk.js
```diff
@@ -797,6 +827,243 @@ function collectNotes(rows) {
...
+    return edge;
+  };
+
+  // Amount-only; PK visibility is decided after, in filterPkGraphByPk.
+  for (const { tx, index, kind } of rows) {
+    const outputs = normalizedOutputs(tx)
+      .filter((slot) => slot.audited && noteMatchesAmount(slot.audited, filters));
+    const inputs = normalizedInputs(tx)
+      .filter((slot) => slot.audited && noteMatchesAmount(slot.audited, filters));
+    const effectiveKind = kind ?? (inputs.length === 0 ? 'deposit' : 'transfer');
```
Fantoni0: Kind is null for viewonly pools, everything will be marked as deposit

Fantoni0 (changes requested):
Last change I promise 🙏

app/js/admin-gvk.js
v0-e: :+1:

Fantoni0 (approved):
LGTM!
Great work! @v0-e

PR #618, opened 2026-09-18
Admin UI simplify allowlist insertion

v0-e:
The user can compute the ASP leaf by themselves, so no need to share both the note public key and ASP secret with the admin.


Fantoni0:
Not sure about this one. Does it really simplify for the average user? Or for ourselves?
Right now the admin panel is just for internal purposes mainly.
image
Public key and ASP secret are secrets that the user is notified about during the onboarding, so they are easy to copy/paste. But if we need to open a console and call the SDK, it seems more complex to me.

On the secrecy of the fields:
- Asp secret only exists for the ASP tree. So it can be shared with the ASP provider.
- Public key is public, and we can assume most forms of whitelist onboarding will require you to identify your pk. 

What do you think?

v0-e:
> Does it really simplify for the average user? Or for ourselves?

For both. 2 values vs 1 value that needs to be shared.

> Public key and ASP secret are secrets that the user is notified about during the onboarding, so they are easy to copy/paste. But if we need to open a console and call the SDK, it seems more complex to me.

This PR also changes the ASP secret provided in the UI to the ASP leaf itself. The ASP secret is now completely abstracted from the user.

> On the secrecy of the fields:

I agree. And the "most forms of whitelist onboarding will require you to identify your pk" is important here :+1:
Though the thing it made me to forego the sharing of the pub key, is that on pools with blocklist-only, the admin can't reliably get to know the pub keys of the users transacting in the pool. So this more or less extends the "privacy" of the pub key to the allowlist pools. Which I agree may not be what we want.

Probably the argument that can be made here against these changes is that the admin will not know the pub key of a an allowed user in case the admin wants to block the user later. Everything else is simplified for both the user and admin.

Fantoni0:
> This PR also changes the ASP secret provided in the UI to the ASP leaf itself. The ASP secret is now completely abstracted from the user.

Missed that, my bad

> Probably the argument that can be made here against these changes is that the admin will not know the pub key of a an allowed user in case the admin wants to block the user later.

Yes, exactly. Each ASP might want to implement this differently. I would say, that since the admin panel is out of the user space, we can close this for now.
We can have a call to discuss it if you feel strongly about this approach instead.

